PR #73, opened 2023-02-06
feat: add support for viewing bookmarked posts

david-hosier:
Same as favourites, but for bookmarks. I only discovered that bookmarks were a thing in Mastodon after I had already created the last PR for favourites, otherwise I would have just done it all at the same time. Sorry for the noise.


davidgarywood:
Separate PRs for individual strands like this are great, as we can just review that one thing at a time 👍

kkostov (approved):
Awesome 🚀

Issue #74, opened 2023-02-07
NSFont is unavailable in Mac Catalyst


TheApApp:
When building for a MacCatalyst app, I receive the following build errors after adding the import and quick start code 
`let instanceURL = URL(string: "my.instance.value")
                let client = TootClient(instanceURL: instanceURL, accessToken: "MYACCESSTOKEN`

The errors show up in NSFont+Traits.swift:
...SourcePackages/checkouts/TootSDK/Sources/TootSDK/Extensions/NSFont+Traits.swift:9:11 'NSFont' is unavailable in Mac Catalyst
this is repeated at lines 9, 11, 15, 19, 28, and 34.

Also receiving the additional errors - see attached image.
<img width="1612" alt="Screenshot 2023-02-06 at 20 17 05" src="https://user-images.githubusercontent.com/2049209/217123237-fd5c2864-5a57-4320-afbe-ceae9271e223.png">

davidgarywood:
@kkostov I _think_ we need to change this to `#if canImport(AppKit) && !targetEnvironment(macCatalyst)`, to ensure this isn't included there

kkostov:
Thank you so much for spotting this @TheApApp, a new version with a fix will follow as soon as the build clears :)

TheApApp:
Thanks for the quick reaction...

PR #75, opened 2023-02-07
Fix/#74 build on mac catalyst

kkostov:
Fixes the build on Catalyst + adds the needed @.available checks to the unit test target, so it can build on tvOS and watchOS


kkostov:
Fixes #74

davidgarywood (approved):
Nice 🧑‍🔧

PR #77, opened 2023-02-07
feat: support signin via ASWebAuthenticationSession on Apple platforms

david-hosier:
Added support for presenting a signin flow via ASWebAuthenticationSession per Discussion #64 . It sounded like Dave was ok considering this as a legit PR with the changes he recommended. I can withdraw if you would rather go a different route. I've also been committing doc updates separately since so many files change as part of that process; I wanted to keep that disentangled from the code changes even though it might not matter that much in the end result.


david-hosier:
Oops, sorry for the linting failure. I'll figure out how to set that up locally and ensure it's good from now on.

davidgarywood:
Thanks for this @david-hosier  - I should be able to review tonight here in NZ, so hopefully have either feedback or an approval within the day 😀

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
```diff
@@ -0,0 +1,46 @@
...
+    ///   - callbackURI: The callback URI  (`redirect_uri`) which was used to initiate the authorization flow. Must match one of the redirect_uris declared during app registration.
+    ///   - prefersEphemeralWebBrowserSession: A Boolean value that indicates whether the session should ask the browser for a private authentication session.
+    ///   - presentationContextProvider: A delegate that provides a display context in which the system can present an authentication session to the user.
+    /// - Returns: The auth token for the user if authentication succeeds.
+    @MainActor public func presentSignIn(callbackURI: String, prefersEphemeralWebBrowserSession: Bool = false, presentationContextProvider: ASWebAuthenticationPresentationContextProviding? = nil) async throws -> String {
+        
+        let authUrl = try await createAuthorizeURL(callbackURI: callbackURI)
+        
+        let returnedUrl: URL = try await withCheckedThrowingContinuation { continuation in
+            let authSession = ASWebAuthenticationSession(url: authUrl, callbackURLScheme: "tootsdk") { (url, error) in
```
Tunous, line 25: callbackURLScheme should match scheme in callbackURI, perhaps you could change callbackURI to URL and extract scheme from here?

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
```diff
@@ -0,0 +1,46 @@
...
+                }
+                if let url = url {
+                    return continuation.resume(returning: url)
+                }
+                return continuation.resume(throwing: TootSDKError.unexpectedError("There was a problem authenticating the user: no URL was returned from the first authentication step."))
+            }
+            
+            if let presentationContextProvider {
+                authSession.presentationContextProvider = presentationContextProvider
+            }
```
Tunous, line 37: Not providing presentation context will fail authentication with error: `Error Domain=com.apple.AuthenticationServices.WebAuthenticationSession Code=2 "Cannot start ASWebAuthenticationSession without providing presentation context. Set presentationContextProvider before calling -start.”`

Because of that it’s better to either make provider parameter required or preferably create default implementation that will be used as fallback.

PR #80, opened 2023-02-08
dev/ASWebAuthentication

davidgarywood:
This adds @Tunous's suggestion, and @david-hosier's great work in adding ASWebAuthentication directly into the client for Apple platforms.

Raising this PR so we can discuss further, with changes I wanted to add.

@kkostov I think this may still need a little finessing, and not least if we're good to go with it then we'll need to update the Quickstart in the README to advise this path!


Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
```diff
@@ -0,0 +1,62 @@
...
+    ///
+    /// - Parameters:
+    ///   - callbackURI: The callback URI  (`redirect_uri`) which was used to initiate the authorization flow. Must match one of the redirect_uris declared during app registration.
+    ///   - prefersEphemeralWebBrowserSession: A Boolean value that indicates whether the session should ask the browser for a private authentication session.
+    ///   - presentationContextProvider: A delegate that provides a display context in which the system can present an authentication session to the user.
+    /// - Returns: The auth token for the user if authentication succeeds.
+    @MainActor public func presentSignIn(callbackURI: String,
+                                         prefersEphemeralWebBrowserSession: Bool = true,
+                                         presentationContextProvider: ASWebAuthenticationPresentationContextProviding? = TootPresentationAnchor()) async throws -> String {
```
Tunous, line 21:
```suggestion
                                         presentationContextProvider: ASWebAuthenticationPresentationContextProviding = TootPresentationAnchor()) async throws -> String {
```

Non-optional so it will be impossible to call this incorrectly. Unless there is a case where nil is valid, but I’m not aware of one.

Tunous:
Also while looking at this code it occurred to me that there might be a lifetime issue of presentationContextProvider parameter. I started a discussion on Swift Forums about that: https://forums.swift.org/t/lifetime-of-object-when-using-withcheckedcontinuation-and-weak-var/63027

But I’m not sure if it’s a real problem here since I used similar approach before and didn’t observe any issues. Or I was just lucky.

david-hosier:
Thanks @Tunous for your input. I see that your Swift Forum post has some feedback that may require some changes here. I'm out of my depth on that conversation since I'm relatively new to Swift.

Thanks @davidgarywood for taking this on.

kkostov (approved):
This looks fantastic folks, good work @david-hosier @Tunous @davidgarywood 🚀

Issue #81, opened 2023-02-09
Detect flavour automatically 


kkostov:
Opening this to track progress on flavour detection.

PR #82, opened 2023-02-10
Detect instance flavour automatically

kkostov:
Adds support for flavour detection #81 

* `TootClient` now has a `connect()` method which retrieves instance information and detects the flavour based on the `version` field.
* Added a new async initializer which internally calls connect, e.g. `let client = try await TootClient(connect: url)`. Uses a different parameter order and label to prevent errors where the sync initializer is used in async context.
* Added a new debug flag `debugInstance` which enables printing of the current instance flavour.
* Added flavour `pixelfed`.
* Extended `swiftyadmin` with a new command `get-flavour`.

Screenshot 2023-02-10 at 03 23 35


davidgarywood (approved):
🔥

PR #84, opened 2023-02-11
Implementation of getFollowers and getFollowing

winsmith:
This PR implements these two APIs:

- https://docs.joinmastodon.org/methods/accounts/#followers
- https://docs.joinmastodon.org/methods/accounts/#following

by adding these two methods to TootClient:

- getFollowers(for id: String)
- getFollowing(for id: String)

I'm not entirely sure that I understand the whole project structure, so I suspect there is more to implementing this API. If something's missing, please let me know :)


kkostov:
@winsmith amazing! Thank you, Daniel! It's perfect 🤩! 

Could you just run `sh ./Scripts/updatedocs.sh` from the root of the library and commit the resulting changes? With that, it will be ready for merging.

winsmith:
I shall do that 😊

winsmith:
Added pagination, docs and rebased on top of the newest main. :)

kkostov (approved):

PR #88, opened 2023-02-13
SwiftUI-Toot Demo: Removed Unused XC Data Model

jnozzi:
Very simple cleanup task. The standard template Core Data model is unused in the SwiftUI-Toot demo. It's been removed.


kkostov (approved):
Awesome! Thank you @jnozzi !

PR #91, opened 2023-02-15
Updated Stream paging + add local/federated calls

davidgarywood:
This PR:

* Fixes our streaming function so that it updates the PageInfo object with the latest results from the API. This means each successive refresh will retrieve only new data.

* Adds Local and Federated timelines GET calls

* Updates the SwiftUI demo app so that it has a segmented picker to switch between home/local/federated timelines


kkostov (approved):
🚀

Issue #92, opened 2023-02-16
Version 0.22 in stream pageInfoparameter is ignored when getting timelines


garryeves:
In the latest version 0.22

    public func stream(_ stream: PostTootStreams, _ pageInfo: PagedInfo? = nil) throws -> AsyncStream<[Post]> 

has been updated to use the new logic of using newHolder?.pageInfo when calling getHomeTimeline.

I am passing in a PageInfo instance and this is not being processed.

it could be that I am doing something wrong, but I am trying to show the toots after a specific id, which I set as minId in the PagedInfo that I then pass when I call the stream when I pass in the PagedInfo as a variable.  The value I pass in is ignored.

Example lines I am using

        let pagingDetails = PagedInfo(minId: "109840050111793127")

        Task {
            for await posts in try await tootManager.currentClient.data.stream(streamType, pagingDetails) {

davidgarywood:
Hi - I’ll take a look at this over the next day, I’ve got a suspicion it’s an issue where the PagedInfo being given, isn’t being used in the first call that’s made.

The _expected_ behaviour is we use the first PagedInfo that’s provided and then set this based on the data returned by the API afterwards.

I’m also wondering if I need to clarify the function to just take a minID rather than the PagedInfo as we’re only using that in the refresh call.

davidgarywood:
Hi @garryeves  - I've a PR up now (#93 ) - which addresses this. If you get a chance to check this before @kkostov reviews, I'd love to hear whether it fixes the issues you've been experiencing or not. It really should do 👍

garryeves:
Thanks.  This is now working really well.

PR #93, opened 2023-02-16
Fix - Use the pageInfo in TootStream when first setting up the stream

davidgarywood:
This is a 1-liner fix; we have just never been using the PageInfo provided, to TootStream 🤦 

This addresses #92


kkostov (approved):
Ah yes, good catch 🚀

PR #95, opened 2023-02-17
changed branch of nuke from master to main

technicat:
nuke has no branch named master so a package reset and subsequent updates break also updated the revision of EmojiText while testing the package update

### DIFF
--- a/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
+++ b/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
@@ -497,8 +497,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/kean/Nuke";
 			requirement = {
-				branch = main;
-				kind = branch;
+				kind = exactVersion;
+				version = 11.6.2;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
--- a/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
+++ b/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
@@ -497,7 +497,7 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/kean/Nuke";
 			requirement = {
-				branch = master;
+				branch = main;
 				kind = branch;
 			};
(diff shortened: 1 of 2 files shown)


Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
```diff
@@ -497,7 +497,7 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/kean/Nuke";
 			requirement = {
-				branch = master;
+				branch = main;
 				kind = branch;
```
kkostov, line 501: I think it's better to pin the version to avoid accidental updates

```suggestion
				kind = exactVersion;
				version = 11.6.2;
```

Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,16 +6,16 @@
       "location" : "https://github.com/divadretlaw/EmojiText",
       "state" : {
         "branch" : "main",
-        "revision" : "510413aeccc50891cf74d6cb6b273125190cdd3c"
+        "revision" : "263935972835482fe655d7ce44e101cc3fb463bf"
```
kkostov, line 9: Let's pin the dependency to the older commit `510413aeccc50891cf74d6cb6b273125190cdd3c` for now. I see EmojiText is making some internal changes that impact layout and caching and should be tested before upgrading.

kkostov (changes requested):
Good catch 🤩! I'm not sure how `master` ended up there, but it's awesome to have this fixed. I made a few small suggestions for tweaking the dependencies, after which it will be ready to merge.

Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
kkostov: I also requested that they tag their versions without the "v" prefix so we can use them as a dependency instead https://github.com/divadretlaw/EmojiText/issues/1

kkostov (approved):

PR #98, opened 2023-02-18
Fix sendable warnings for Xcode 14.3 (beta)

davidgarywood:
This fixes the sendable warnings invoked with Xcode 14.3.

It's entirely possible I've missed some with this PR, and would definitely appreciate some 👀 !


Tunous (approved):

kkostov (approved):
I think it looks good for now!

PR #100, opened 2023-03-02
Add support for hashtag timelines and only_media timeline filter

daprice:
This adds support for the `/api/v1/timelines/tag/:hashtag` endpoint including the parameters to add or exclude additional tags. For my use case I also needed to use the `only_media` parameter, so I went ahead and added support for that to the local and federated timelines as well, hopefully that’s ok but I can break it out as a separate PR if needed.

For local/remote filtering I added the `TimelineLocality` enum which I thought seemed more swifty than the separate `local` and `remote` params that the Mastodon API provides (unless I’m misunderstanding something, I don’t think it makes any sense to be able to make a request filtered by local-only *and* remote-only at the same time).

To support all this in `TootStream`, I added the extra parameters as associated values on the relevant enum cases because that made the most sense to me, but let me know if you’d prefer a different way of handling them.

I also added a demo of this to the sample app – you can tap a hashtag in the search results to view the timeline for it.

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Request.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Request.swift
@@ -16,7 +16,7 @@ extension TootClient {
         return url
     }
     
-    internal func getQueryParams(_ pageInfo: PagedInfo? = nil, limit: Int? = nil, offset: Int? = nil) -> [URLQueryItem] {
+    internal func getQueryParams(_ pageInfo: PagedInfo? = nil, limit: Int? = nil, offset: Int? = nil, onlyMedia: Bool? = nil, locality: TimelineLocality? = nil) -> [URLQueryItem] {
         var queryParameters = [URLQueryItem]()
         
         if let maxId = pageInfo?.maxId {
@@ -37,6 +37,16 @@ extension TootClient {
 
         if let offset = offset {
             queryParameters.append(.init(name: "offset", value: String(offset)))
+        }
+        
+        if onlyMedia == true {
+            queryParameters.append(.init(name: "only_media", value: "true"))
+        }
+        
+        if locality == .local {
+            queryParameters.append(.init(name: "local", value: "true"))
+        } else if locality == .remote {
+            queryParameters.append(.init(name: "remote", value: "true"))
         }
         return queryParameters
     }
--- a/Sources/TootSDK/TootClient/TootClient+Request.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Request.swift
@@ -39,15 +39,24 @@ extension TootClient {
             queryParameters.append(.init(name: "offset", value: String(offset)))
         }
         
-        if onlyMedia == true {
-            queryParameters.append(.init(name: "only_media", value: "true"))
+        if let onlyMedia = onlyMedia {
+            queryParameters.append(.init(name: "only_media", value: String(onlyMedia)))
         }
         
-        if locality == .local {
-            queryParameters.append(.init(name: "local", value: "true"))
-        } else if locality == .remote {
-            queryParameters.append(.init(name: "remote", value: "true"))
+        if let locality = locality {
+            switch locality {
+            case .local:
+                queryParameters.append(.init(name: "local", value: "true"))
+            case .remote:
+                queryParameters.append(.init(name: "remote", value: "true"))
+            case .all:
+                queryParameters.append(contentsOf: [
+                    .init(name: "local", value: "false"),
+                    .init(name: "remote", value: "false")
+                ])
+            }
         }
+        
         return queryParameters
     }
 
--- a/Examples/swiftui-toot/TootSDK-Demo/Views/Feed/FeedSelectionView.swift
+++ b/Examples/swiftui-toot/TootSDK-Demo/Views/Feed/FeedSelectionView.swift
@@ -18,8 +18,8 @@ struct FeedSelectionView: View {
     @State private var selection: SelectionOptions = .home
     
     @StateObject var timeLineHomeViewModel = FeedViewModel(streamType: .timeLineHome)
-    @StateObject var timeLineLocalViewModel = FeedViewModel(streamType: .timeLineLocal)
-    @StateObject var timeLineFederatedViewModel = FeedViewModel(streamType: .timeLineFederated)
+    @StateObject var timeLineLocalViewModel = FeedViewModel(streamType: .timeLineLocal(onlyMedia: nil))
+    @StateObject var timeLineFederatedViewModel = FeedViewModel(streamType: .timeLineFederated(onlyMedia: nil))
     
     var body: some View {
         VStack {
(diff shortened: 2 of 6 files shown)


davidgarywood:
Hi Dale - I'll be able to give this a proper look over the weekend, but thank you this looks like some great additions!

davidgarywood (changes requested):
This adds functionality that I'm really keen for us to have; both hashtag calls and only_media are things I want to play with too! So thank you 😀

I've a few things I think could be amended here to pull it together just a little more, and leave us in a good state for other parameters etc in the future.

Create a new struct, called something like `PostQueryParams` where all these things can live. We can then migrate the onlyMedia flag, arrays of any/all/none hashtags, to that, along with separate booleans for local and remote.

This'll make it nicer when passing along from TootStreams to the get calls internally as well, and easier to extend later.

I'd like us to keep PageInfo and limit as they are though as their duties are slightly different to just parameters and control of that is needed for moving up/down a timeline etc.

We should also be able to create the streams without passing parameters, just passing the enum along, so we don't have `.timeLineFederated()` but can just use `.timeLineFederated`. `PostQueryParams` can be an optional parameter that defaults to nil.

So, a little bit of a refactor, but like I said it should give us some flexibility for future things! 🙏

Sources/TootSDK/TootClient/TootClient+Request.swift
```diff
@@ -38,6 +38,16 @@ extension TootClient {
         if let offset = offset {
             queryParameters.append(.init(name: "offset", value: String(offset)))
         }
+        
+        if onlyMedia == true {
+            queryParameters.append(.init(name: "only_media", value: "true"))
+        }
```
Tunous, line 44: Unlikely, but if default value changed it would be impossible to pass false. I suggest always adding this param when it's not nil.

daprice:
> Create a new struct, called something like `PostQueryParams` where all these things can live. We can then migrate the onlyMedia flag, arrays of any/all/none hashtags, to that, along with separate booleans for local and remote.
> 
> This'll make it nicer when passing along from TootStreams to the get calls internally as well, and easier to extend later.

Should there be anything to prevent using the wrong parameters for the timeline type then? E.g. someone requesting the local timeline but passing a `PostQueryParams` containing the any/all/none hashtag parameters?


> We should also be able to create the streams without passing parameters, just passing the enum along, so we don't have `.timeLineFederated()` but can just use `.timeLineFederated`. `PostQueryParams` can be an optional parameter that defaults to nil.

As I understand it, if an enum case has associated values, Swift requires the parentheses when instantiating it, even if all the associated values have defaults/are optional. I’m not sure where else to store the params other than an associated value because TootStream uses the case as the key for caching, and should be able to hold separate caches for, say, the timeline for a tag with and without onlyMedia for example.

Sources/TootSDK/TootClient/TootClient+Request.swift
daprice: Good catch, I’ll fix that there and also for the local/remote filtering.

davidgarywood:
The more I think about all of this, I'm wondering if we need a different angle over all. You're right in the issue of the dictionary too.

Key points:

* We don't want to have to keep passing every param as individual values all the way along either from the enum to the get functions.
* We still want the caching
* We want default query params for each type of timeline

So here's what I think e can do.

* Turn PostTootStreams into an enum that's just PostTootStreamType, with each case having NO params attached to it.
* Create a new struct for PostQueryParams, with default parameters, and use this with our get calls for posts.
* Add a function inside of the new PostTootStreams to create the default parameters for its case.

We can then have a new struct for the TootStreams:

```
struct StreamQuery: Hashable {
   var timeline: PostTootStreamType
   var queryParams: PostQueryParams
}
```

Which can then be used in this way:

```
let query = StreamQuery(.home, PostQueryParams(…))

and then it’s .refresh(streamQuery) etc
```

The dictionary can then use the StreamQuery struct instead of the enum as it's key

@kkostov  keen if you have any more thoughts here. but otherwise I think this is the direction this should take

kkostov:
@davidgarywood I love that idea, I think it will be very neat. The main guideline is to minimize friction when a developer is using this API and this proposal covers it.

daprice:
One possible issue I see with this proposal: the hashtag timeline requires at least the “base” hashtag to be in the query. If all parameters are in a separate struct, then someone using the API could go:

```
let query = StreamQuery(.hashtag, PostQueryParams(onlyMedia = true))
```

…or something like that, leaving out the mandatory hashtag. There would need to be additional logic to check that the `PostQueryParams` value meets the requirements for the timeline type. (Where would this logic go? Failable initializer for `StreamQuery`? Throw an error when making the actual request? Just let it make the query and get a 404 from the server? All of these sound annoying to deal with if I’m trying to use the library)

daprice:
What about:

1. keep the parameters as associated values of `PostTootStreams` cases (`.timeLineLocal(onlyMedia:)`, `.timeLineTag(tag:, anyTags:, allTags…)`, and so on).
2. To avoid having to pass all the individual parameters down the entire chain, refactor the separate timeline getting methods to a single method like `TootClient.getTimeline(_ query: PostTootStreams, _ pageInfo: PagedInfo? = nil, limit: Int? = nil) -> PagedResult<[Post]>` that takes the entire enum case as a parameter rather than separate parameters for each query option
3. also refactor `TootClient.getQueryParams()` to take the entire enum case as a parameter (leave the paging info as separate parameters as it is now)
4. rename `PostTootStreams` to reflect its new role (suggestions: `TimelineType`, `TimelineSpecifier`, `TimelineQuery`, `StreamQuery`, or keep your suggestion of `PostTootStreamType`)

This would (I think) address all of your key points of avoiding passing individual parameters all the way, keeping `PostTootStreams` (or whatever it ends up being called) as the caching key, and having defaults for each type (in the form of default values for the associated values). It would also preserve the compiler enforcement of using only the correct set of parameters for each timeline type because the timeline-specific parameters would be coupled to that type.

It would be a breaking change for anything that uses the timeline getting methods directly, though. And timelines that have the onlyMedia option would still need the empty trailing parentheses if someone wants the default value for that parameter.

Tunous:
> And timelines that have the onlyMedia option would still need the empty trailing parentheses if someone wants the default value for that parameter

Or additionally there can be extension property that will call corresponding enum case:

```swift
extension PostTootStreams {
    public static var timeLineLocal: PostTootStreams {
        return .timeLineLocal()
    }
}
```

davidgarywood:
Righty, so me and @kkostov had a bit of a chat earlier on about this, and I want to avoid PR table tennis if we can!

So... I'd like to change the scope of this PR to just the TootClient+Request calls, and remove the TootStream additions. 

We can then move discussion of how to include these calls in TootStream, to our [GitHub discussions area](https://github.com/TootSDK/TootSDK/discussions).

This issue was raised yesterday: https://github.com/TootSDK/TootSDK/issues/103, with regards to Mastodon's streaming API. Ideally I'd like to think about the scope of what TootStream does, potentially including this to enhance its behaviour.

I realise this might not leave you feeling too pleased, but I think it'll let us get the getHashTag call in and reach somewhere really good for TootStream itself :+1:

kkostov:
Indeed, something straightforward like `TootClient.getHashtagTimeline(query: HashtagTimelineQuery, _ pageInfo: PagedInfo? = nil, limit: Int? = nil)` would be enough for now.

daprice:
The cleanest way for me to isolate just the TootClient+Request changes was to create a new branch from main and cherry pick those commits into it, so I had to make a new PR for it: #105. Closing this one.

PR #101, opened 2023-03-05
API to update media attachment

Tunous:
https://docs.joinmastodon.org/methods/media/#update
https://api.pleroma.social/#operation/MediaController.update

Because mime type is only necessary when updating thumbnail I decided to provide 2 separate initializers for params. First skips thumbnail and second makes it required together with mime type. It might be more inconvenient if someone would want to update thumbnail only conditionally, but we are preventing situation where someone passes it without mime type.


kkostov (approved):
Awesome @Tunous! Thank you 🤩!

PR #102, opened 2023-03-05
[BREAKING] Remove Attachment type in favour of MediaAttachment

kkostov:
While reviewing #101, I noticed we have a duplication of types - `Attachment` and `MediaAttachment` were used in the context of attachment. 
* Remove `Attachment` in favour of the type name also used by the Mastodon API.
* Remove the use of a nested type alias for attachment identifiers (Attachment.ID -> String).


Tunous:
If I remember correctly I added this duplicated type. It is made on purpose. There is one crucial difference between them. MediaAttachment has optional url which might be missing if media is not yet uploaded.

kkostov:
@Tunous Nice, I've totally missed that before. 

I think the difference between the two types was too subtle to notice unless one is already familiar with how the Mastodon REST API is working. This is also specific behaviour for Mastodon (other server types handle it differently).

I added a comment to clarify why `url` can be nil - Do you think this is enough?

Tunous:
I would prefer to not have this optionality but I guess it is a good approach for the sake of simplicity. Just a single additional filtering out of unwanted values in my case.

Tunous (approved):

kkostov:
@Tunous what do you mean when you say filtering - can posts be listed with an optional url?

Tunous:
I don't think that can happen but with optional url in this type I either have to handle optional url everywhere, use force unwrapping, or ignore values with nil. I will be using the last approach. Essentially filtering out values where it will be nil before mapping to type from UI layer that has it required.

kkostov:
You're right, it's not very cool to have the URL as an optional. How about this, then:

* `MediaAttachment.url` becomes non-optional (this is the type used by Posts and most commonly seen in other API calls).
* `func uploadMedia(_ params: [UploadMediaAttachmentParams], mimeType: [String]) async throws -> [MediaAttachment]` becomes `func uploadMedia(_ params: [UploadMediaAttachmentParams], mimeType: [String]) async throws -> [String]` returning just the identifiers of the uploaded attachments. This is the value needed when using the attachments to create a post.
* To check if a post has finished uploading, one would need to call `func getMedia(id: [String]) async throws -> [MediaAttachment]?`

Tunous:
I like that except for only id in response to `uploadMedia`. Having a way to tell whether the media has finished uploading immediately removes the need to call `getMedia`. It will only be necessary for unfinished uploads. (Faster final response, lower api usage, fewer chances for error due to network issues)

In my app I use type with id and state. But perhaps instead of state optional url might be useful for others. Although same could be said for other values and we slowly reach what Attachment has :)

```swift
public struct UploadedMedia: Identifiable, Sendable {
    public let id: String
    public let state: State

    public enum State: Sendable {
        case inProgress
        case succeeded
    }
}
```

kkostov:
Awesome! Then I'd amend the change to the following:

* `MediaAttachment.url` becomes non-optional (this is the type used by Posts and most commonly seen in other API calls).
* `func uploadMedia(_ params: [UploadMediaAttachmentParams], mimeType: [String]) async throws -> [MediaAttachment]` becomes `func uploadMedia(_ params: [UploadMediaAttachmentParams], mimeType: [String]) async throws -> [UploadedMedia]`
* To check if a post has finished uploading, one would need to call `func getMedia(id: [String]) async throws -> [MediaAttachment]?`

davidgarywood (approved):
Love the collaboration on this - thanks @Tunous for your input here!

PR #105, opened 2023-03-07
TootClient+Request support for hashtag timeline and only_media

daprice:
Submitting this as a new PR with the requested scope change from #100 because it was easiest for me to create a new branch and cherry pick the relevant commits.

As discussed in the other PR thread, this adds hashtag timeline and only_media support to TootClient only (leaving TootStream alone) in the form of `getHashtagTimeline(query: HashtagTimelineQuery, _ pageInfo: PagedInfo? = nil, limit: Int? = nil)`, with the various tag-related parameters moved to the `HashtagTimelineQuery` struct.

I left `onlyMedia` as an optional (default nil) parameter on `getLocalTimeline()` and `getFederatedTimeline()` – since that was the only extra parameter for those timelines, I didn’t see any value in breaking it out into a query struct.


Sources/TootSDK/TootClient/TootClient+TimeLine.swift
```diff
@@ -41,12 +41,13 @@ public extension TootClient {
     /// - Parameters:
     ///   - pageInfo: a PageInfo struct that tells the API how to page the response, typically with a minId set of the highest id you last saw
     ///   - limit: Maximum number of results to return (defaults to 20 on Mastodon with a max of 40)
+    ///   - onlyMedia: Return only statuses with media attachments
     /// - Returns: a PagedResult containing the posts retrieved
-    func getLocalTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
+    func getLocalTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil, onlyMedia: Bool? = nil) async throws -> PagedResult<[Post]> {
```
kkostov, line 46: Even though it's a single, field, let's move this to a `LocalTimelineQuery` struct as a first optional parameter:
 ```suggestion
    func getLocalTimeline(_ query: LocalTimelineQuery? = nil, _ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```

File: Sources/TootSDK/TootClient/TootClient+TimeLine.swift [hunk 2]
```diff
@@ -55,12 +56,39 @@ public extension TootClient {
     /// - Parameters:
     ///   - pageInfo: a PageInfo struct that tells the API how to page the response, typically with a minId set of the highest id you last saw
     ///   - limit: Maximum number of results to return (defaults to 20 on Mastodon with a max of 40)
+    ///   - onlyMedia: Return only statuses with media attachments
     /// - Returns: a PagedResult containing the posts retrieved
-    func getFederatedTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
+    func getFederatedTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil, onlyMedia: Bool? = nil) async throws -> PagedResult<[Post]> {
```
kkostov, line 61: Even though it's a single, field, let's move this to a `FederatedTimelineQuery` struct as a first optional parameter:
```suggestion
    func getFederatedTimeline(_ query: FederatedTimelineQuery? = nil, _ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```

Sources/TootSDK/TootClient/TootClient+TimeLine.swift
```diff
@@ -55,12 +56,39 @@ public extension TootClient {
...
+        return try await getPosts(req, pageInfo, limit)
+    }
+    
+    /// Retrieves public statuses containing the given hashtag
+    /// - Parameters:
+    ///   - query: a ``HashtagTimelineQuery`` struct defining the parameters of the request
+    ///   - pageInfo: a PageInfo struct that tells the API how to page the response, typically with a minId set of the highest id you last saw
+    ///   - limit: Maximum number of results to return (defaults to 20 on Mastodon with a max of 40)
+    /// - Returns: a PagedResult containing the posts retrieved
+    func getHashtagTimeline(query: HashtagTimelineQuery, _ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```
kkostov, line 76: To keep this consistent with other timeline requests, let's formulate it as:
```suggestion
    func getHashtagTimeline(_ query: HashtagTimelineQuery, _ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```

Sources/TootSDK/HTTP/TimelineLocality.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  File.swift
+//
+//  Created by Dale Price on 3/1/23.
+//
+
+import Foundation
+
+/// Specifies whether a timeline request should be restricted to local or remote statuses
+public enum TimelineLocality: Codable, Sendable {
```
kkostov, line 10: The notion of "locality" if I understand what you mean by it is related to a Post, not to a timeline (is the post local or is the post remote). There are a number of timelines that do not have a clear indication if they include only local or remote posts or both, and so we'd need "unknown/undefined" as a case to the enum which feels confusing.

Let's remove the notion of `locality` in favour of additional filter fields in the query structs of each timeline, and only for the timelines that have support for them (like `only_media`).

Sources/TootSDK/HTTP/HashtagTimelineQuery.swift
```diff
@@ -0,0 +1,30 @@
...
+//  HashtagTimelineQuery.swift
+//  
+//
+//  Created by Dale Price on 3/7/23.
+//
+
+import Foundation
+
+/// Specifies the parameters for a hashtag timeline request
+public struct HashtagTimelineQuery: Codable, Sendable {
```
kkostov, line 11: Could you include a public initializer with all fields and a perhaps a second initializer taking a single string tag:

```swift
HashtagTimelineQuery(tag: "", anyTags:...)
HashtagTimelineQuery(tag: "")
```

Sources/TootSDK/HTTP/TimelineLocality.swift
daprice: It referred to where a timeline endpoint should filter posts from, but if it isn’t clear then it’s probably better to stick to separate `local` and `remote` booleans like Mastodon has. Done in 0b1ff77

Sources/TootSDK/HTTP/HashtagTimelineQuery.swift
daprice: Whoops, forgot the default initializer is `internal`. Fixed in aeb4a0c

daprice:
@kkostov Thanks so much for the feedback, I think I got everything.

kkostov (approved):
🎉

PR #109, opened 2023-03-10
Make it possible to register an account

kkostov:
This PR makes it possible to register for new accounts. It can be tested with the provided example:
```shell
swiftyadmin register-account -u https://someinstance -n <username> -e “<your email>“ -p “<your password>“
```

* Adds `registerAccount` method
* Extends login flow to allow for client credentials grant
* Updates README with instructions on how the create an account
* Adds a register account example to swiftyadmin
* Some housekeeping tweaks to timelines


Sources/TootSDK/TootClient/TootClient+TimeLine.swift
```diff
@@ -52,6 +52,15 @@ public extension TootClient {
         return try await getPosts(req, pageInfo, limit)
     }
     
+    /// Retrieves the user's local timeline
+    /// - Parameters:
+    ///   - pageInfo: a PageInfo struct that tells the API how to page the response, typically with a minId set of the highest id you last saw
+    ///   - limit: Maximum number of results to return (defaults to 20 on Mastodon with a max of 40)
+    /// - Returns: a PagedResult containing the posts retrieved
+    func getLocalTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```
Tunous, line 60: How is this method different from the one above? `query` is already optional in the other one too. Won't there be conflicts if someone passes no parameters?

Sources/TootSDK/TootClient/TootClient+TimeLine.swift
```diff
@@ -67,6 +76,15 @@ public extension TootClient {
         return try await getPosts(req, pageInfo, limit)
     }
     
+    /// Retrieves the user's federated timeline
+    /// - Parameters:
+    ///   - pageInfo: a PageInfo struct that tells the API how to page the response, typically with a minId set of the highest id you last saw
+    ///   - limit: Maximum number of results to return (defaults to 20 on Mastodon with a max of 40)
+    /// - Returns: a PagedResult containing the posts retrieved
+    func getFederatedTimeline(_ pageInfo: PagedInfo? = nil, limit: Int? = nil) async throws -> PagedResult<[Post]> {
```
Tunous, line 84: Same question as above

Sources/TootSDK/TootClient/TootClient+TimeLine.swift
kkostov: A previous PR introduced something that caused these signatures to appear with leading nil at the call site:
```swift
return try await getFederatedTimeline(nil, pageInfo, limit: limit)
```

So, this change adds an overload so they remain available without the leading `nil`.

Tunous: @kkostov I incorrectly assumed that it would be possible to call original function without nil as first parameter. I see now that is not possible because swift gives error about incorrect type.

About the error/warning part, I didn't see how swift would decide which function will be called since call site matched both. But looks like it has some rules that I don't know.

Tunous (approved):

Sources/TootSDK/Models/LocalTimelineQuery.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  File.swift
```
davidgarywood: ~~Tiny little nit - let's delete these `File.swift` header comments!~~ - have gone after them myself :D

davidgarywood (approved):

PR #110, opened 2023-03-10
implement Identifiable where id getter is defined

technicat:
Implemented Identifiable where an id getter is defined (found 17 cases, 6 of them already implement Identifiable)

See https://github.com/TootSDK/TootSDK/discussions/108


daprice:
Isn’t it a security vulnerability to use ID strings provided by servers for this?

It could also be an issue if you’re developing a client that supports multiple accounts, as different servers seem to use a different ID string for the same post. I doubt there would normally be a collision where different items from different servers have the same ID, but a malicious server could intentionally cause an ID collision.

davidgarywood:
A malicious server probably could, or at least issue the same id twice in a set of data. I think that’s an edge case I’d be happy to acknowledge as existing here but not requiring us to counter it in the SDK.

If you’re connecting multiple TootClients to multiple instances, you need to keep that data separate anyway. Or wrap up in some way to represent where it’s from. I’m doing this at the moment with the client app I’m working on by using a separate cache in Boutique for each instance+user on that the user connects.

daprice:
That’s true, just feels wrong to give the server any influence on an app’s logic beyond merely populating content (i.e. hooking into everything identifiable could be used for). But maybe it’s not as big of a problem as I’m imagining

Sources/TootSDK/Models/DomainBlock.swift
```diff
@@ -18,7 +18,7 @@ public enum DomainBlockSeverity: String, Codable, Hashable {
 }
 
 /// Represents a domain limited from federating.
-public struct DomainBlock: Codable, Hashable {
+public struct DomainBlock: Codable, Hashable, Identifiable {
```
kkostov, line 21: This one I would keep as not `identifiable` and even make the `id` optional as it's not provided supported by all servers.

kkostov:
Thanks a lot for this PR @technicat, there is just a minor tweak needed for `DomainBlock` before I can merge it.

Sources/TootSDK/Models/DomainBlock.swift
kkostov, line 21:
```suggestion
public struct DomainBlock: Codable, Hashable {
```

Sources/TootSDK/Models/DomainBlock.swift
```diff
@@ -18,7 +18,7 @@ public enum DomainBlockSeverity: String, Codable, Hashable {
...
     /// The ID of the DomainBlock in the database.
     public var id: String
```
kkostov, line 23:
```suggestion
    public var id: String?
```

kkostov (approved):
awesome 🎉

PR #113, opened 2023-03-12
Toot Stream Updates

davidgarywood:
This is a draft PR for discussion over at https://github.com/TootSDK/TootSDK/discussions/104

Specifically this addresses:

✅  How to handle errors of misconfigured parameters if we used generic ones (we check the type of query params supplied and return an error if the wrong type is used)

✅  How to ensure call sites are 'clean'. e.g if you're running on defaults it can look something like: client.data.stream(.timeLineHome)

✅ Ensuring that the TootStream private dictionary where it caches requested streams, has a unique key for the stream type + params.

This does not cover:
_How would we like the Mastodon streaming API to interact with TootStream?_ (I figure this can be a separate PR after this one is discussed/merged!)

⚠️  🚨 Please only add to this PR's conversation with code review items. Any discussion of approach, please reply over at: https://github.com/TootSDK/TootSDK/discussions/104


kkostov (approved):
🤩🚀

Tunous (approved):

PR #125, opened 2023-03-16
delete/favourite/unfavouritePost return Post instead of Post?

technicat:
Code should never return nil (either returns Post or throws) so make it consistent with the other functions like boost.

https://github.com/TootSDK/TootSDK/discussions/124


kkostov (approved):
Nice! Thank you! @technicat

PR #128, opened 2023-03-18
Comment fix in List.swift

technicat:
repliesPolicy had the same comment as title, replaced it with the doc from the Mastodon API spec

Also removed an extra space in ListRepliesPolicy comments.


davidgarywood (approved):

davidgarywood:
Thanks for this!

PR #135, opened 2023-03-24
Fixes issue with minId being set to nil when no content is returned

### LINKED ISSUE #134: TootDataStream.refresh causing duplicate content for home feed
Hi, thanks for all the great work you are doing on this, it's really helpful.
I spotted an issue with refresh which is hard to spot in the SwiftUI-Toot demo but seems to happen consistently, where `min_id` isn't being consistently added to the timelines home request, seems to work as expected for local and federated though.  

Steps to recreate:
Open the SwiftUI-Toot example project
Set client to debugRequests: `self.currentClient?.debugRequests = true`
Launch the project and pull to refresh the home feed in the Browse feed tab
You should see something like:
➡️ 🌏 GET https://mastodon.social/api/v1/timelines/home?min_id=110073176170390408
then pull to refresh again
➡️ 🌏 GET https://mastodon.social/api/v1/timelines/home?
The min_id is no longer set

If you scroll down on the feed you will also see the duplicate content

andrewconlan:
Closes #134

### DIFF
--- a/Sources/TootSDK/TootStream/TootStream+Posts.swift
+++ b/Sources/TootSDK/TootStream/TootStream+Posts.swift
@@ -12,7 +12,7 @@ extension TootDataStream {
     
     /// Provides an async stream of updates for the given stream
     /// - Parameters:
-    ///   - timeline: the type of post stream to updarte
+    ///   - timeline: the type of post stream to update
     ///   - pageInfo: PagedInfo object for max/min/since ids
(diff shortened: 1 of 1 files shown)


davidgarywood (approved):
Thank you @andrewconlan - very much appreciated! 🙏

kkostov (approved):
Very neat! Thank you 🙏!

PR #122, opened 2023-03-15
User posts timeline

Tunous:
https://api.pleroma.social/#operation/AccountController.statuses
https://docs.joinmastodon.org/methods/accounts/#statuses

PR #140, opened 2023-04-01
Dev/add account posts

davidgarywood:
This adds a getAccountPosts call so that we can retrieve an account's posts


Tunous:
Isn't this covering the same api as https://github.com/TootSDK/TootSDK/pull/122 ?

davidgarywood:
> Isn't this covering the same api as #122 ?

It is! I'd managed to miss it as it was over in the timeline extension and not account 🤦  

I'll close this PR - thanks @Tunous

PR #144, opened 2023-04-10
Get trending tags, posts, and links

daprice:
This PR adds `getTrendingTags(limit:offset:)`, `getTrendingPosts(limit:offset:)` and `getTrendingLinks(limit:offset:)` methods to TootClient.

For the links, the format returned by the Masto API is similar to the existing structure for link preview cards, but not quite the same, so I created a new struct `TrendingLink` for it. This endpoint does use the same format for usage history as trending tags and tag search, so I renamed the existing `TagHistory` struct to `UsageHistory` so I could use it in `TrendingLink`.

Support for the trending endpoints appears to be inconsistent between the instance flavors that TootSDK supports, with some of their documentation being a little bit vague about what specific endpoints are supported or, in the case of the trending tags endpoint, using a different URL path for what appears to be the same endpoint. I *think* I handled these correctly, but I don’t have an account on every flavor of instance to test on.


kkostov:
Thanks! This looks awesome! It's great to get support for trends. Here are a few points we could try to iterate on:

* For flavours besides mastodon, the function can return an empty array without calling the server.
* `UsageHistory` is a bit vague as a name. Perhaps we can try to use something closer to the name provided by the API e.g. `History`? It can be a nested type declared in each of the trending entities, e.g. `Tag.History` and `TrendingLink.History` as there is no guarantee that these are always the same types (and we don't have examples from other flavours presently).
* This is just nice to have, but it would be wonderful if it could be added to one of the examples, e.g. Swifty admin where new command can be added by copying existing ones. This way we can quickly test it against several servers.

kkostov (changes requested):
Thanks! This looks awesome! It's great to get support for trends. Here are a few points we could try to iterate on:

* For flavours besides mastodon, the function can return an empty array without calling the server.

* `UsageHistory` is a bit vague as a name. Perhaps we can try to use something closer to the name provided by the API e.g. `History`? It can be a nested type declared in each of the trending entities, e.g. `Tag.History` and `TrendingLink.History` as there is no guarantee that these are always the same types (and we don't have examples from other flavours presently).

* This is just nice to have, but it would be wonderful if it could be added to one of the examples, e.g. Swifty admin where new command can be added by copying existing ones. This way we can quickly test it against several servers.

daprice:
Huh, on further testing it looks like Friendica has full support for trending links and posts as well (based on instances that the directory says are using the stable version), even though their [documentation](https://wiki.friendi.ca/docs/api-mastodon) says otherwise. Do you think I should enable Friendica support for those too?

Examples/swiftyadmin/Sources/swiftyadmin/Trends/GetTrendingTags.swift
```diff
@@ -0,0 +1,33 @@
...
+    
+    @Option(name: .shortAndLong, help: "Maximum number of results to return")
+    var limit: Int?
+    
+    @Option(name: .shortAndLong, help: "Skip the first n results")
+    var offset: Int?
+    
+    mutating func run() async throws {
+        print("Listing trending tags")
+        let client = TootClient(instanceURL: URL(string: url)!, accessToken: token)
```
kkostov, line 21:
```suggestion
        let client = try await TootClient(connect: URL(string: url)!, accessToken: token)
```

Examples/swiftyadmin/Sources/swiftyadmin/Trends/GetTrendingPosts.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    @Option(name: .shortAndLong, help: "Maximum number of results to return")
+    var limit: Int?
+    
+    @Option(name: .shortAndLong, help: "Skip the first n results")
+    var offset: Int?
+    
+    mutating func run() async throws {
+        print("Listing trending posts")
+        let client = TootClient(instanceURL: URL(string: url)!, accessToken: token)
```
kkostov, line 21:
```suggestion
        let client = try await TootClient(connect: URL(string: url)!, accessToken: token)
```

Examples/swiftyadmin/Sources/swiftyadmin/Trends/GetTrendingLinks.swift
```diff
@@ -0,0 +1,29 @@
...
+	
+	@Option(name: .shortAndLong, help: "Maximum number of results to return")
+	var limit: Int?
+	
+	@Option(name: .shortAndLong, help: "Skip the first n results")
+	var offset: Int?
+	
+	mutating func run() async throws {
+		print("Listing trending links")
+		let client = TootClient(instanceURL: URL(string: url)!, accessToken: token)
```
kkostov, line 21:
```suggestion
		let client = try await TootClient(connect: URL(string: url)!, accessToken: token)
```

kkostov (approved):
Nicely done @daprice, looks perfect! Thank you so much!

PR #145, opened 2023-04-11
PostEdit implements Hashable

technicat:
Seems like many or most of the TootSDK models implement Hashable, but PostEdit doesn't and in this case I need PostEdit to be Hashable to pass to a custom NavigationStack path like this:

```swift
enum Route : Hashable {
    case postDetail(Post)
    case postEdit(PostEdit)
    case profile(Account)
    case reply(Post)
}
```


davidgarywood (approved):
🙏

kkostov (approved):
🎉

PR #149, opened 2023-04-21
Report and ScheduledPost implement Identifiable

technicat:
missed these two in the previous PR

https://github.com/TootSDK/TootSDK/pull/110


kkostov (approved):
Ah indeed, good catch!

PR #150, opened 2023-04-21
API to get, follow and unfollow tags

Tunous:


davidgarywood (approved):
Brilliant 🙏

Sources/TootSDK/TootClient/TootClient+Tags.swift
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        return try await fetch(Tag.self, req)
+    }
+
+    /// Follow a tag.
+    /// - Parameter id: Name of the tag.
+    @discardableResult
+    func followTag(_ id: String) async throws -> Tag {
+        let req = HTTPRequestBuilder {
```
kkostov, line 25: This method is supported only by Mastodon and Friendica, we need a guard for the flavour.

Sources/TootSDK/TootClient/TootClient+Tags.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        return try await fetch(Tag.self, req)
+    }
+
+    /// Unfollow a tag.
+    /// - Parameter id: Name of the tag.
+    @discardableResult
+    func unfollowTag(_ id: String) async throws -> Tag {
+        let req = HTTPRequestBuilder {
+            $0.url = getURL(["api", "v1", "tags", id, "unfollow"])
```
kkostov, line 38: Same here, only masto and Friendica

kkostov:
@Tunous Looks great! If you could you add a guard for the flavour of the follow/unfollow methods then it will be ready to merge 🚀.

Thanks a lot! 🙌

Tunous:
@kkostov Good find. This also gives me an idea that I should hide that function from the app UI. I added a property to tell if the feature is supported to be able to do that.

kkostov (approved):
@Tunous Nice! I like the availability flag a lot! (@davidgarywood)

PR #153, opened 2023-04-25
Fix release notes not tagging me correctly 😇

kkostov:
hardcoded the handle for GitHub API does not return the correct value


davidgarywood (approved):
*shakes fist at Github* 😅

PR #155, opened 2023-04-29
added followed_tags access

technicat:
Added TootClient.getFollowedTags() to TootClient+Tags.swift

https://docs.joinmastodon.org/methods/followed_tags/


technicat:
Btw, also tested in my tootsdk app (probably should add something to the tootsdk example app).

Simulator Screenshot - iPhone 14 - 2023-04-29 at 13 41 23

kkostov (approved):
Awesome! Thank you!

PR #160, opened 2023-05-06
Add macOS sandbox entitlement instructions to the README

MrRooni:
## Reason to Be

When starting a new macOS project I immediately ran into a sandbox entitlement error. I thought it might be helpful to be explicit in the quick start section about enabling the correct network entitlement.

## Thought Process

I decided to add this as a disclosed section to try and keep the README as visually clean as possible.


kkostov (approved):
Good catch! Thank you! And I love the screenshot with the cool arrow :)!

PR #162, opened 2023-05-11
cleanup: refactor fetch of paged result into a shared function

technicat:
added TootClient.fetchPagedResult(req) called by all the functions that return PagedResult.

It's basically the same as getPosts (which I changed to call this new function), but with a generic type (specified as Decodable, but that can be omitted, the type inference handles it).


davidgarywood (approved):
Nice use of turning something repeated into something generic - thank you! 🙏

kkostov (approved):
It's nice to get rid of all that boilerplate. Thank you, @technicat 🙌!

PR #163, opened 2023-05-18
feat: view follow suggestions

technicat:
TootClient.getSuggestions() performs this query

https://docs.joinmastodon.org/methods/suggestions/

### DIFF
--- a/Sources/TootSDK/Models/Suggestion.swift
+++ b/Sources/TootSDK/Models/Suggestion.swift
@@ -1,0 +1,35 @@
+//
+//  Suggestion.swift
+//  
+//
+//  Created by Philip Chu on 5/17/23.
+//
+
+import Foundation
+
+/// An account suggested by the server.
+public struct Suggestion: Codable, Hashable, Sendable {
+    public init(source: Source, account: Account) {
+        self.source = source
+        self.account = account
+    }
+
+    /// The reason this account is being suggested.
+    public var source: Source
+    /// The account being recommended to follow.
+    public var account: Account
+
+    public enum Source: String, Codable, Sendable {
+        /// This account was manually recommended by your administration team
+        case staff
+        /// You have interacted with this account previously
+        case past_interactions
+        /// This account has many reblogs, favourites, and active local followers within the last 30 days
+        case global
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case source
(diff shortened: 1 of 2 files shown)


Sources/TootSDK/Models/Suggestion.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// The reason this account is being suggested.
+    public var source: Source
+    /// The account being recommended to follow.
+    public var account: Account
+
+    public enum Source: String, Codable, Sendable {
+        /// This account was manually recommended by your administration team
+        case staff
+        /// You have interacted with this account previously
+        case past_interactions
```
Tunous, line 26:
```suggestion
        case pastInteractions = "past_interactions"
```

Sources/TootSDK/Models/Suggestion.swift
```diff
@@ -0,0 +1,35 @@
...
+        /// This account has many reblogs, favourites, and active local followers within the last 30 days
+        case global
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case source
+        case account
+    }
```
Tunous, line 34: Not needed, it's the same as automatically generated one.

```suggestion
```

Sources/TootSDK/Models/Suggestion.swift
davidgarywood: Do we need this? (The decoder is set to handle camel case coming in)

Tunous: You're right, fixed my suggestion.

davidgarywood: @technicat just to say this is a great addition - if we can update this case to snake case, then I think this is good to go! 🙏

technicat: Went with the first version of @Tunous suggestion, got a decoder error with the second. Here's how it looks now on universeodon.com

IMG_2387

davidgarywood (approved):
🙌

Sources/TootSDK/Models/Suggestion.swift
Tunous: Ah, camel case conversion works for fields not enum values. That's why my first suggestion worked :)

PR #167, opened 2023-05-19
Improve compatibility for web push notifications and pagination

kkostov:
* [all platforms] Adds the `reblog` alert from the push notification request.
* [all platforms] Removes the identifier from `PushSubscription` (which is unused for API calls) as it requires special handling because it can be a `String` (Pleroma/Akkoma) or an `Int` (Mastodon) depending on server type.
* [all platforms] Enhances pagination to tolerate spaces and line breaks before or after pagination header values (Pleroma/Akkoma seems to produce them sometimes when paginating notifications).
* [Linux] Pagination link parsing now uses [WebURL](https://github.com/karwa/swift-url) to offer consistent validation across platforms. (cf. [see here](https://github.com/TootSDK/TootSDK/pull/167#issuecomment-1554916164))

### DIFF
--- a/Sources/TootSDK/HTTP/Pagination.swift
+++ b/Sources/TootSDK/HTTP/Pagination.swift
@@ -2,6 +2,7 @@
 // Copyright (c) 2022. All rights reserved.
 
 import Foundation
+import WebURL
 
 public struct Pagination {
     public var maxId: String?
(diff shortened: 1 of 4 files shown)


davidgarywood (approved):
Oof @ Linux Swift not working the same ! Looking good 🙏

PR #168, opened 2023-05-20
add removeSuggestion

technicat:
Added TootClient.removeSuggestion which accesses this API

https://docs.joinmastodon.org/methods/suggestions/#remove

I tested this on universeodon.com and it works, but apparently only on suggestions with a past_interactions source (which isn't documented but kind of makes sense, that's the only suggestion type that is tailored to the user)

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Account.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Account.swift
@@ -31,7 +31,10 @@ extension TootClient {
     }
 
     /// Get all accounts which follow the given account, if network is not hidden by the account owner.
-    /// - Parameter id: the ID of the Account in the instance database.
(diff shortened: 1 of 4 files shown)


Sources/TootSDK/TootClient/TootClient+Account.swift
```diff
@@ -44,7 +47,10 @@ extension TootClient {
     }
 
     /// Get all accounts which the given account is following, if network is not hidden by the account owner.
-    /// - Parameter id: the ID of the Account in the instance database.
+    /// - Parameters:
+    ///     - id: the ID of the Account in the instance database.
```
kkostov, line 51: Nice! 🙌

kkostov (approved):

PR #171, opened 2023-06-10
feat: featured_tags API

technicat:
Added TootClient functions to access the Mastodon featured_tags API for featuring/unfeaturing tags, and viewing and suggesting featured tags.

https://docs.joinmastodon.org/methods/featured_tags/

Placed them in TootClient+FeaturedTags.swift and moved the existing getFeaturedTags (that takes an account ID) there.

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+FeaturedTags.swift
+++ b/Sources/TootSDK/TootClient/TootClient+FeaturedTags.swift
@@ -7,14 +7,25 @@
 
 import Foundation
 
-extension TootClient {
+public extension TootClient {
     /// Get tags featured by user.
     ///
     /// - Parameter userID: ID of user in database.
     /// - Returns: The featured tags or an error if unable to retrieve.
-    public func getFeaturedTags(forUser userID: String) async throws -> [FeaturedTag] {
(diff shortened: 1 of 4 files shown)


Sources/TootSDK/TootClient/TootClient+FeaturedTags.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+        return try await fetch([FeaturedTag].self, req)
+    }
+
+    /// Shows up to 10 recently-used tags.
+    ///
+    /// - Returns: Array of ``Tag``.
+    func getFeaturedTagsSuggestions() async throws -> [Tag] {
+        let req = HTTPRequestBuilder {
+            $0.url = getURL(["api", "v1", "featured_tags", "suggestions"])
```
kkostov, line 39: All endpoints using `featured_tags` should be limited to the mastodon flavour.

kkostov (changes requested):
Looks good!
I just peeked into the docs of other flavours and it seems these are all mastodon-only.

kkostov (approved):
Nice! Thank you!

davidgarywood (approved):
Great stuff @technicat - thank you!

PR #172, opened 2023-06-11
Log debug messages using OSLog logger

Tunous:
Xcode 15 introduced improvements to debug console with extra support for OSLog messages. This pull request is just a quick change to start a discussion. Feel free to expand on it or share ideas. I didn’t test how this works on non iOS targets.

Some ideas for changes that could be made:
- Separate logger with different category for request/response instead of emoji
- Flag to disable logging of headers

Issues to resolve:
- It seems like there is a limit to length of displayed messages in console. Long messages appear cut. Not sure if this is customizable somewhere, it's a bug in Xcode, there is a different way to log long messages or it's entirely not supported.
- Check how it behaves on earlier versions of Xcode.

Examples:

Screenshot 2023-06-10 at 20 22 33
Screenshot 2023-06-10 at 20 26 01
Screenshot 2023-06-10 at 20 26 19

Reference:

https://developer.apple.com/wwdc23/10226
https://developer.apple.com/documentation/os/logging

### DIFF
--- a/Sources/TootSDK/Models/TootSDKFlavour.swift
+++ b/Sources/TootSDK/Models/TootSDKFlavour.swift
@@ -3,7 +3,7 @@
 
 import Foundation
 
-public enum TootSDKFlavour: String, Codable, Sendable, CaseIterable {
+public enum TootSDKFlavour: String, Codable, Sendable, CaseIterable, CustomStringConvertible {
     /// Mastodon server. API Documentation can be found at https://docs.joinmastodon.org/api/
     case mastodon
 
@@ -18,4 +18,8 @@ public enum TootSDKFlavour: String, Codable, Sendable, CaseIterable {
 
     /// Akkoma server. API Documentation can be found at https://docs.akkoma.dev/stable/development/API/differences_in_mastoapi_responses/
     case akkoma
+
+    public var description: String {
+        rawValue
+    }
 }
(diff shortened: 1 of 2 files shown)


kkostov:
Yes, that's a good idea. I've been trying out the new Console in Xcode 15 beta and indeed the emojis are no longer helpful now that we have a more powerful console view.
I would simplify the logging as you suggest:
* introduce a category (let's keep it simple for now, we have 2 categories—request and response?)
* remove all emojis
* consider grouping some logs into a single output (e.g. the headers) so they also appear in a single "segment" in the console in Xcode (less noise)

Tunous:
About problem with message length I found in [release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Known-Issues) that it's a known issue. It might be resolved in future beta:

> The maximum permitted size of os_log messages shown in the debug console is smaller than in previous Xcode versions. This may cause long messages to more commonly appear truncated in the debug console. (109381234)
> 
> Workaround: Configure the Run scheme action in your project to set the environment variable IDELogRedirectionPolicy to the value oslogToStdio to stream logs via standard IO with increased size limits. This policy disables structured metadata for os logs.

kkostov:
@Tunous  is this still something you'd like to continue exploring or perhaps we can close the PR for now? Once we approach release of this Xcode version, we could see what would be the best way to tackle it. What do you think?

Tunous:
Ok, let's close this for now. From my experiments it's not worth using until they resolve the issue with length limit.

PR #177, opened 2023-07-20
API for update_credentials

technicat:
added updateCredentials in TootClient+Account, takes arguments in UpdateCredentialsParams
added PATCH method to support the call
patterned the multipart encoding/aggregation after uploadMedia (update_credentials accepts uploads of avatar/header images), and added a couple of MultipartPart convenience inits to avoid repeatedly pasting header text
SwiftNIO and SwiftSoup packages are updated


davidgarywood (approved):
Excellent - thank you for this @technicat !

PR #180, opened 2023-08-12
workaround for ScheduledPost.sensitive returning as string instead of bool

technicat:
Scheduling a post with a content warning was breaking reading of scheduled posts because the sensitive flag was returned as a string instead of bool. Workaround is to have a custom decoder that calls decodeFromBool (patterned after the existing decodeFromInt).

Other misc stuff: fix comment doc for scheduledAt, make ScheduledPost Hashable (so it can be passed in routes), and update packages, specifically swift-nio.

### DIFF
--- a/Sources/TootSDK/Models/ScheduledPost.swift
+++ b/Sources/TootSDK/Models/ScheduledPost.swift
@@ -4,10 +4,10 @@
 @preconcurrency import struct Foundation.Date
 
 /// Represents a post that will be published at a future scheduled date.
-public struct ScheduledPost: Codable, Identifiable, Sendable {
+public struct ScheduledPost: Codable, Equatable, Hashable, Identifiable, Sendable {
     /// ID of the scheduled post in the database.
     public var id: String
-    /// ID of the post in the database.
+    /// The timestamp for when the status will be posted.
     public var scheduledAt: Date?
     /// The parameters to be used when the post is posted
     public var params: ScheduledPostParams
--- a/Sources/TootSDK/Models/ScheduledPostParams.swift
+++ b/Sources/TootSDK/Models/ScheduledPostParams.swift
@@ -8,7 +8,7 @@
 @preconcurrency import struct Foundation.Date
 
 /// Parameters to post a new scheduled post
-public struct ScheduledPostParams: Codable, Sendable {
+public struct ScheduledPostParams: Codable, Equatable, Hashable, Sendable {
 
     ///  Creates parameters to create a new scheduled post
     /// - Parameters:
@@ -77,5 +77,22 @@ public struct ScheduledPostParams: Codable, Sendable {
         case scheduledAt = "scheduled_at"
         case contentType = "content_type"
         case inReplyToConversationId = "in_reply_to_conversation_id"
+    }
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.text = try? container.decode(String.self, forKey: .text)
+        self.mediaIds = try? container.decode([String].self, forKey: .mediaIds)
+        self.poll = try? container.decode(CreatePoll.self, forKey: .poll)
+        self.inReplyToId = try? container.decode(String.self, forKey: .inReplyToId)
+        // Mastodon incorrectly returns this as string
+        self.sensitive = try? container.decodeBoolFromString(forKey: .sensitive)
+        self.spoilerText = try? container.decode(String.self, forKey: .spoilerText)
+        self.visibility = (try? container.decode(Post.Visibility.self, forKey: .visibility)) ?? .public
+        self.language = try? container.decode(String.self, forKey: .language)
+        self.idempotency = try? container.decode(String.self, forKey: .idempotency)
+        self.scheduledAt = try? container.decode(Date.self, forKey: .scheduledAt)
+        self.contentType = try? container.decode(String.self, forKey: .contentType)
+        self.inReplyToConversationId = try? container.decode(String.self, forKey: .inReplyToConversationId)
     }
 }
--- a/Sources/TootSDK/Models/ScheduledPostParams.swift
+++ b/Sources/TootSDK/Models/ScheduledPostParams.swift
@@ -74,9 +74,9 @@ public struct ScheduledPostParams: Codable, Equatable, Hashable, Sendable {
         case visibility
         case language
         case idempotency
-        case scheduledAt = "scheduled_at"
-        case contentType = "content_type"
-        case inReplyToConversationId = "in_reply_to_conversation_id"
+        case scheduledAt
+        case contentType
+        case inReplyToConversationId
     }
     
     public init(from decoder: Decoder) throws {
--- a/Sources/TootSDK/Models/ScheduledPostParams.swift
+++ b/Sources/TootSDK/Models/ScheduledPostParams.swift
@@ -88,7 +88,7 @@ public struct ScheduledPostParams: Codable, Equatable, Hashable, Sendable {
         // Mastodon incorrectly returns this as string
         self.sensitive = try? container.decodeBoolFromString(forKey: .sensitive)
         self.spoilerText = try? container.decodeIfPresent(String.self, forKey: .spoilerText)
-        self.visibility = (try? container.decode(Post.Visibility.self, forKey: .visibility)) ?? .public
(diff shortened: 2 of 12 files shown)


kkostov:
@technicat do you have an example json payload with the boolean-as-string? It will be nice to have a unit test for this, so we make sure we didn't forget a property for the custom decoder. I can add the unit test if you give me write access to the branch.

Sources/TootSDK/Models/ScheduledPostParams.swift
```diff
@@ -78,4 +78,21 @@ public struct ScheduledPostParams: Codable, Sendable {
...
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.text = try? container.decode(String.self, forKey: .text)
+        self.mediaIds = try? container.decode([String].self, forKey: .mediaIds)
+        self.poll = try? container.decode(CreatePoll.self, forKey: .poll)
+        self.inReplyToId = try? container.decode(String.self, forKey: .inReplyToId)
+        // Mastodon incorrectly returns this as string
+        self.sensitive = try? container.decodeBoolFromString(forKey: .sensitive)
+        self.spoilerText = try? container.decode(String.self, forKey: .spoilerText)
```
kkostov, line 90: For all optional fields, we should use `decodeIfPresent` as they may not be available in the payload.

Sources/TootSDK/Models/ScheduledPost.swift
```diff
@@ -4,10 +4,10 @@
 @preconcurrency import struct Foundation.Date
 
 /// Represents a post that will be published at a future scheduled date.
-public struct ScheduledPost: Codable, Identifiable, Sendable {
+public struct ScheduledPost: Codable, Equatable, Hashable, Identifiable, Sendable {
     /// ID of the scheduled post in the database.
     public var id: String
-    /// ID of the post in the database.
+    /// The timestamp for when the status will be posted.
```
kkostov, line 10: ❤️ much better

technicat:
> 

Here it is from my console (also reported on https://github.com/mastodon/mastodon/issues/24645 )

[
  {
    "id" : "1023",
    "scheduled_at" : "2023-08-31T08:12:00.000Z",
    "params" : {
      "scheduled_at" : null,
      "in_reply_to_id" : null,
      "media_ids" : null,
      "application_id" : 91595,
      "poll" : null,
      "spoiler_text" : "Warn",
      "idempotency" : null,
      "allowed_mentions" : null,
      "text" : "Test",
      "language" : "en",
      "visibility" : "direct",
      "with_rate_limit" : false,
      "sensitive" : "true"
    },
    "media_attachments" : [

    ]
  }
]

Sources/TootSDK/Models/ScheduledPostParams.swift
technicat: Here it is from my console (also reported on https://github.com/mastodon/mastodon/issues/24645 )
```json
[
{
"id" : "1023",
"scheduled_at" : "2023-08-31T08:12:00.000Z",
"params" : {
"scheduled_at" : null,
"in_reply_to_id" : null,
"media_ids" : null,
"application_id" : 91595,
"poll" : null,
"spoiler_text" : "Warn",
"idempotency" : null,
"allowed_mentions" : null,
"text" : "Test",
"language" : "en",
"visibility" : "direct",
"with_rate_limit" : false,
"sensitive" : "true"
},
"media_attachments" : [

]

}
]
```

technicat: [
{
"id" : "1023",
"scheduled_at" : "2023-08-31T08:12:00.000Z",
"params" : {
"scheduled_at" : null,
"in_reply_to_id" : null,
"media_ids" : null,
"application_id" : 91595,
"poll" : null,
"spoiler_text" : "Warn",
"idempotency" : null,
"allowed_mentions" : null,
"text" : "Test",
"language" : "en",
"visibility" : "direct",
"with_rate_limit" : false,
"sensitive" : "true"
},
"media_attachments" : [

]

}
]

technicat: Added decodeIfPresent to ScheduledPostParams and also for one property in FeaturedTag

technicat: added scheduled_status_sensitive.json to the test resources

kkostov:
@technicat added tests and pushed to your branch, feel free to check them out.

The test revealed a problem - `TootDecoder` is not picking all keys in the container with `keyDecodingStrategy` set to `convertFromSnakeCase` as it's used in the rest of the library. 

I'm not sure what's causing this - if you have any idea that would be awesome.

technicat:
> 

I think it's because of the snakecase mappings in the CodingKeys. If I remove "spoiler_text" from there or change it to "spoilerText" the test passes.

technicat:
I removed the snakecase mappings from CodingKeys in ScheduledPostParams which were getting in the way of the decoder automatic snakecase conversion and also changed inReplyToId to Int as in the spec and actual return values (although it's properly a String when passed into the schedule request). Also added more test cases to check for reply and attachment cases.

technicat:
Here are the changes reflected in my rudimentary scheduled post view. The clip and arrow icons indicate replies and attachments.

IMG_3834

Sources/TootSDK/Models/ScheduledPostParams.swift
```diff
@@ -58,24 +58,41 @@ public struct ScheduledPostParams: Codable, Sendable {
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.text = try? container.decodeIfPresent(String.self, forKey: .text)
+        self.mediaIds = try? container.decodeIfPresent([String].self, forKey: .mediaIds)
+        self.poll = try? container.decodeIfPresent(CreatePoll.self, forKey: .poll)
+        self.inReplyToId = try? container.decodeIfPresent(Int.self, forKey: .inReplyToId)
+        // Mastodon incorrectly returns this as string
+        self.sensitive = try? container.decodeBoolFromString(forKey: .sensitive)
+        self.spoilerText = try? container.decodeIfPresent(String.self, forKey: .spoilerText)
+        self.visibility = (try? container.decode(Post.Visibility.self, forKey: .visibility)) ?? .public
```
kkostov, line 91: Unknown visibility should trigger a decoding error:

```suggestion
        self.visibility = try container.decode(Post.Visibility.self, forKey: .visibility)
```

kkostov:
@technicat thanks for looking into it. That really makes more sense. Looking into the `schedulePost` method, we actually send this as multipart and receive it as JSON, so we do need:

* CodingKeys that correspond to the signature expected by the server for the `ScheduledPostParams` object (using the snake case overrides) - it's the `FormDataEncoder` which will convert the object to the expected form data fields for the request.

* A separate struct to describe the ScheduledPost detail (e.g.`ScheduledPostDetail`), which will be the response of the server after creating a scheduled post (or listing them). This one will use the `TootDecoder` when parsing the incoming json data.

So in `ScheduledPost.swift` we need to replace `public var params: ScheduledPostParams` with `public var params: ScheduledPostDetail`

* `in_reply_to_id` is not always an int, for example `Pleroma/Akkoma` flavours always return a string for post identifiers cf

kkostov:
Yes - `SheduledPostParams` is used both params and detail, let's keep it as is.

technicat:
> Yes - `SheduledPostParams` is used both params and detail, let's keep it as is.
> 
> Edit: Our "convention" is to call the request objects provided to Toot methods Params (what gets passed into `schedulePost`), so then I guess `ScheduledPostRequest` can be renamed to `SheduledPostParams` and `SheduledPostParams` to `SheduledPost..Detail?`
> 
> But I'm OK with keeping it as is as well. Let me know if it's ready to merge @technicat.

I'll do one more commit to change inReplyToID back to String (this will avoid breaking any existing callers, too).

I like the renaming idea, I did find the current names confusing, too! Maybe next PR? Also, I noticed applicationID was missing from ScheduledPostParams but didn't seem like a pressing issue (I don't know where you can look up the IDs).

technicat:
OK, I changed inReplyToID back to String. Ready to merge!

davidgarywood (approved):
Nice fix!

PR #181, opened 2023-08-12
Add isAnonymous property

Tunous:


Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -340,4 +340,9 @@ extension TootClient {
         }
         self.flavour = instance.flavour
     }
+    
+    /// Returns `true` if this instance of `TootClient` has an `accessToken`.
+    public var isAuthorized: Bool {
```
kkostov, line 345: I think this field may be promising more than intended. Even if there is a cached access token, it doesn't mean it's a working token. The consuming app is better suited to perform this check with the correct intention as it will be more aware of the lifecycle of TootClient.

The best way to check if the client can be used for API calls is to invoke `verifyCredentials()` - if the method succeeds and returns the user's Account, then the current token is valid.

Tunous: Updated naming and implementation to better match my usage. I want to detect if a client is fully anonymous.

kkostov (approved):

Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -28,14 +28,14 @@ public class TootClient: @unchecked Sendable {
     public lazy var data = TootDataStream(client: self)
     /// The clientName the client was initialized with
     public let clientName: String
+    /// The current accessToken in use
+    public internal(set) var accessToken: String?
```
davidgarywood, line 32: Honestly, I don't think this is a good route to go. The consuming app has responsibility for persisting the token (in the keychain please!) and giving it to the TootClient, so you should already have the current token stored securely and shouldn't need to read it back from the client.

One of the reasons the accessToken is pushed down and made internal, is also to protect it from being easily read by any 3rd party services outside of TootSDK. It is an edge-case, but it's one we don't have to expose.

TLDR: I don't mind the isAnonymous flag, but I'd rather we leave this property internal to the SDK.

davidgarywood (changes requested):

Sources/TootSDK/TootClient/TootClient.swift
Tunous: Reverted changes leaving only `isAnonymous`.

davidgarywood (approved):
🙏

PR #185, opened 2023-08-19
added TootClient.getLists(id) to get lists an account is in

technicat:
also added a List decode test

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Lists.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Lists.swift
@@ -14,6 +14,16 @@ public extension TootClient {
 
         return try await fetch([List].self, req)
     }
+    
+    /// User lists that you have added this account to.
+    func getLists(id: String) async throws -> [List] {
+        let req = HTTPRequestBuilder {
+            $0.url = getURL(["api", "v1", id, "lists"])
+            $0.method = .get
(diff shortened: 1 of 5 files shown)


Sources/TootSDK/TootClient/TootClient+Lists.swift
```diff
@@ -14,6 +14,16 @@ public extension TootClient {
 
         return try await fetch([List].self, req)
     }
+    
+    /// User lists that you have added this account to.
+    func getLists(id: String) async throws -> [List] {
```
Tunous, line 19: This name doesn't fit to what the function does. Maybe try something like these to make it clear?

```suggestion
    func getContainingLists(accountId: String) async throws -> [List] {
```

```suggestion
    func getListsContainingAccount(id: String) async throws -> [List] {
```

kkostov (approved):

Sources/TootSDK/TootClient/TootClient+Lists.swift
kkostov: Oh I totally missed that, you're right!

kkostov: PR https://github.com/TootSDK/TootSDK/pull/186

PR #186, opened 2023-08-19
fix: rename getLists to getListsContainingAccount

kkostov:
This improves the getList implementation from #185 with a name that better describes its function and also gets moved to the Account related methods.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,3 +5,4 @@ disabled_rules:
   - line_length
   - trailing_whitespace
   - unused_optional_binding
+  - line_length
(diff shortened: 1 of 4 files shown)


kkostov:
@Tunous @technicat hey folks, this PR applies the last suggestions after adding the list method https://docs.joinmastodon.org/methods/accounts/#lists

Tunous (approved):

davidgarywood (approved):
👍 prefer this naming!

kkostov:
Thanks for catching this @Tunous 🚀

Issue #27, opened 2022-12-21
ROADMAP - Application behaviour detection


davidgarywood:
* Convenience function to detect application behaviours such as editing posts

Enum (or collection of Enums TBD), to represent actions such as editing posts. e.g `TootAction.editPost`

When the tootclient is connected, the allowed actions are created based on the instance type and version. We do this manually, based on knowledge of what instance/version supports what action. This creates an array on the client, that stores the TootAction enums that are available.

A convenience method for checking if something is available will be created too (this literally just checks our array contains whatever is being queried and returns true/false).

We can then also use that same convenience method within our action's methods, to early exit and return an error.

A new TootSDK error of UnsupportedAction will be created, and will return the action that is not supported within it too.

PR #188, opened 2023-08-19
API to represent features and check if they are supported

Tunous:
This pull request builds upon #187. Now that there are more features that I would like to check if are supported I decided to unify the api.

- Added a new `TootFeature` feature that describes requirements for a feature. Right now it has a list of supported flavours but could be extended to other requirements in future.
- Added requireFeature internal method that can be used to perform runtime check where necessary
- Use this new API to add featured tags and hashtag following features.
- This is a breaking change as it removes previously used `canXXX` variables. If necessary this removal can be reversed to preserve API compatibility.

Possibly related to #27.

### DIFF
--- a/Sources/TootSDK/Models/TootFeature.swift
+++ b/Sources/TootSDK/Models/TootFeature.swift
@@ -1,0 +1,8 @@
+import Foundation
+
+/// Represents a feature that is not supported by all flavours.
+public struct TootFeature {
+
+    /// Flavours that support this feature.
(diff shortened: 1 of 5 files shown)


kkostov (approved):
Nice! I think this is a good way to express the concept of features in a more managed way while allowing the client to check if a specific functionality is available.

davidgarywood (approved):
This is great - love it 🙌

PR #191, opened 2023-08-22
feat: TootClient.setNoteForAccount (sets private note)

technicat:
Accesses this endpoint: https://docs.joinmastodon.org/methods/accounts/#note

Added decode tests for FeaturedTag, Tag, and Relationship

Updated packages, specifically swift-nio.


kkostov (approved):

davidgarywood (approved):
Nice 👍

PR #192, opened 2023-09-02
feat: TootClient.getFamiliarFollowers

technicat:
https://docs.joinmastodon.org/methods/accounts/#familiar_followers


kkostov (approved):

kkostov (changes requested):
@technicat what is the reason for the changes in [98954fb](https://github.com/TootSDK/TootSDK/pull/192/commits/98954fb03b21a6e18476996439628c5e260baf48)?

If a consuming app requires instance information beyond what's provided by a connected instance of `TootClient`, they can invoke the `getInstanceInfo()` method.

technicat:
> @technicat what is the reason for the changes in [98954fb](https://github.com/TootSDK/TootSDK/pull/192/commits/98954fb03b21a6e18476996439628c5e260baf48)?
> 
> If a consuming app requires instance information beyond what's provided by a connected instance of `TootClient`, they can invoke the `getInstanceInfo()` method.

The reason for the change is so I don't have to call getInstanceInfo again, e.g. for use in my Instance display view. connect only calls getInstanceInfo to set the flavour property in TootClient and then discards it so it's kind of a waste. What I've discussed with @Tunous on Mastodon is that the flavour property could be replaced by an instance property, so you can always access the flavour and anything else in the instance, but this is a quicker intermediate change so anyone can save the instance after calling creating the TootClient. Another approach that would enforce more consistency is have getInstanceInfo set the flavour or instance property, which is essentially what I'm doing in a wrapper function (which I call instead of connect)

```swift
func getInstance() async throws -> Instance {
        let instance = try await client.getInstanceInfo()
        if client.debugInstance {
            print("🎨 Detected fediverse instance flavour: \(instance.flavour), version: \(instance.version)")
        }
        client.flavour = instance.flavour
        self.instance = instance
        return instance
    }
```

kkostov:
I think it will be better if this change goes into a separate PR or a discussion here on GitHub. I didn’t see your discussion on fedi (feel free to link) so I may be missing some context but I see some challenges with this approach:

- We aim for TootClient instances to be as close to stateless as possible, it should not retain server information that can become stale nor require a refresh mechanism - all these are concerns for the consuming app to implement in a way that matches its lifecycle and expectations.

- The implementation of `connect` is private and  subject to change (from the point of view of a consuming app). So yes we do need to get instance info today, but that doesn’t mean we don’t want to make a more clever way of detecting the flavour. The connect API should not promise more than it’s intended function.

An extra call to get instance information does not seem wasteful - it’s something a consuming app may want to do frequently in order to obtain the latest configurations (e.g. toot limits etc)

davidgarywood:
Agreed re: not “crossing the streams” in PRsas much as possible 👍

Appreciate the revert @technicat, let’s take the broader points to a GitHub discussion? 🙏

kkostov:
@technicat I just noticed the example JSON for the unit test includes a real person's profile information, would it be possible to use tootsdk's account as an example instead? 

I just saw we have this in other JSON resources as well, I will make a separate PR to clean them up as well.

technicat:
> @technicat I just noticed the example JSON for the unit test includes a real person's profile information, would it be possible to use tootsdk's account as an example instead?
> 
> I just saw we have this in other JSON resources as well, I will make a separate PR to clean them up as well.

That is actually my familiar followers view of TootSDK. Do you mean replace that follower in that JSON with a copy and paste of TootSDK account JSON? I just removed the test for now.

Simulator Screenshot - iPhone 14 - 2023-09-07 at 08 55 41

cblanos:
That screenshot kinda jumped out at me 😄

kkostov (approved):

PR #196, opened 2023-09-12
Add lists feature

Tunous:
Lists are not available on Pixelfed and always return empty response.

https://github.com/pixelfed/pixelfed/blob/e36d7da8413e700edcd9b639db7c44b2b97394fc/app/Http/Controllers/Api/ApiV1Controller.php#L1558


davidgarywood (approved):
🙏

technicat:
The code comment says

/// - Warning: Not available for Pleroma.

Is that supposed to be Pixelfed?

Tunous:
@technicat Oops, you're right. Fixed in #197.

PR #198, opened 2023-09-13
TootClient get/delete filters and a pixelfed fix for updateCredentials

technicat:
Updated Filter from v1 to v2, added TootClient.getFilter, getFilters, and deleteFilter.
Also added a pixelfed case for updateCredentials, they use POST instead of PATCH.

9E86C77D-4E47-4482-A883-D1046D1D2F56


kkostov:
Nice! 
The v2 endpoint is not available for Pleroma and Akkoma. Can we introduce something like `V2.Filters` feature similar to `lists` where the supported flavours are listed?

technicat:
> Nice! The v2 endpoint is not available for Pleroma and Akkoma. Can we introduce something like `V2.Filters` feature similar to `lists` where the supported flavours are listed? We will need v1 support for other flavours and older versions of Mastodon.

I added .filtersV2 (V2.Filters doesn't compile) only available on Mastodon

davidgarywood (approved):

kkostov (approved):
Nice! Let's see how this works. I like that we're trying out a way to version the flavour guards.

technicat:
> Nice! Let's see how this works. I like that we're trying out a way to version the flavour guards.

Could do the same with search. I didn't add a feature for that because I'm not sure what to do with the conditionals around the current requireFlavour.

PR #201, opened 2023-09-23
updated platform features

technicat:
added feature definitions for endorsements, trends, updateCredentials, domainBlocks, removeFollowers, privateNote, familiarFollowers, scheduledPost, and suggestions.

also added pixelfed to the hashtagFollowing feature (the actual following doesn't seem to work well, but it's in the official user interface and the followed tags show up when you query them)


kkostov:
Nice!
- I would like to avoid having to update the comment every time e.g. `Available for Mastodon, Akkoma, and Pleroma.`.
- Apply a `swiftlint lint --fix --format`.

technicat:
> Nice!
> 
>     * I would like to avoid having to update the comment every time e.g. `Available for Mastodon, Akkoma, and Pleroma.`. Not sure if there is a better way to make this visible without peeking into the source code? @davidgarywood ?
> 
>     * Apply a `swiftlint lint --fix --format`.

Reformatted and removed all the "Available" comments. I saw @Tunous did that recently with the lists feature.

kkostov (approved):

PR #205, opened 2023-09-27
Parse instance configuration from instance info request

Tunous:
https://docs.joinmastodon.org/entities/V1_Instance/#configuration

Tested for 
- https://mastodon.world
- https://pxlmo.com
- https://poa.st
- https://venera.social
- https://blob.cat


Tunous:
I see that SwiftLint failed due to type nesting level. Should I move these types or reconfigure/disable this rule?

kkostov:
Please avoid nested types.

kkostov:
oh, I see this is something we’ve had as an issue before. Let’s disable the rule for this PR. It can be fixed separately

davidgarywood (approved):

kkostov (approved):

PR #206, opened 2023-10-01
Allow accessing details about unprocessed media attachments

Tunous:
I'm adding video upload support to my app. In order to make user experience better for not yet uploaded videos I want to display their previews as soon as they are available. Right now this is impossible because TootSDK returns information other than processing state only when upload fully finishes.

In order to implement this I made following changes:

1. Changed `UploadedMediaAttachment` returned by `uploadMedia(_:mimeType:)` call to be an enum with both possible states and related models available as associated value.
2. Introduced `MediaAttachmentProtocol` with information that API returns both for fully uploaded and not yet processed media. This protocol is accessible as `mediaAttachment` property on `UploadedMediaAttachment`
3. Introduced `UnprocessedMediaAttachment` that has the same properties as `MediaAttachment` minus `url`.
4. Depending on the presence of `url` property in API response made TootSDK decode proper model.

If someone used uploadMedia before these changes they don't have to change their code. Both state and id that were previously returned are still available as properties of `UploadedMediaAttachment`.

### DIFF
--- a/Sources/TootSDK/Models/MediaAttachment.swift
+++ b/Sources/TootSDK/Models/MediaAttachment.swift
@@ -4,7 +4,7 @@
 import Foundation
 
 /// Represents a file or media attachment that can be added to a post.
-public struct MediaAttachment: Codable, Hashable, Identifiable, Sendable {
+public struct MediaAttachment: Codable, Hashable, Identifiable, Sendable, MediaAttachmentProtocol {
 
     public init(id: String, type: AttachmentType, url: String, remoteUrl: String? = nil, previewUrl: String? = nil, meta: AttachmentMeta? = nil, description: String? = nil, blurhash: String? = nil) {
         self.id = id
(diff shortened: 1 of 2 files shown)


davidgarywood:
A couple of thoughts / questions on this:

* What stops an app from showing a local preview of the media the user is uploading, regardless of the state of that upload?

* Does this work the same for PixelFed / Akkoma etc as well as Mastodon?

Tunous:
> * What stops an app from showing a local preview of the media the user is uploading, regardless of the state of that upload?

Nothing, sounds like a reasonable thing to do.

> * Does this work the same for PixelFed / Akkoma etc as well as Mastodon?

In their case upload finishes before they send response response. You'll get the model with url.

davidgarywood:
I guess I'm thinking;  the dev has the original media, so you can display a preview/thumbnail etc based on that instead of anything the server gives back. If the other server types don't give anything back until the upload finishes then we'd be adding this just for Mastodon.

With these things in mind, I'm not sure this change is necessary?  (I could easily be missing something though!)

Tunous:
Yes you're right. I changed my implementation based on your suggestion and indeed do not need anything more than what was previously available.

I'm leaving this up to you to decide whether to merge. Can't see other use cases now, but perhaps there is something we can't see.

kkostov:
Nice, thank you for the detailed review @Tunous. I think it's best that we close this for now. When we have a specific case to tackle we can revisit the idea.

davidgarywood:
Yup thank you @Tunous - and I always appreciate your input to the SDK! It’s fine to propose these things even if we end up reaching a different choice in the course of the PR 👍

PR #207, opened 2023-10-03
Make UIKitAttribStringRenderer macOS Compatible

FahimF:
Changed UIKitAttribStringRenderer so that it compiles under macOS (not Catalyst) correctly. Also changed the example project so that it will build as a macOS target.

Maybe would want to rename UIKitAttribStringRenderer since it's not just for UIKit?

### DIFF
--- a/Examples/swiftui-toot/SwiftUI-Toot-Info.plist
+++ b/Examples/swiftui-toot/SwiftUI-Toot-Info.plist
@@ -15,12 +15,5 @@
 			</array>
 		</dict>
 	</array>
-	<key>UIApplicationSceneManifest</key>
-	<dict>
-		<key>UIApplicationSupportsMultipleScenes</key>
-		<true/>
-		<key>UISceneConfigurations</key>
-		<dict/>
-	</dict>
 </dict>
 </plist>
--- a/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
+++ b/Examples/swiftui-toot/SwiftUI-Toot.xcodeproj/project.pbxproj
@@ -405,6 +405,7 @@
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "SwiftUI-Toot-Info.plist";
+				INFOPLIST_KEY_UIApplicationSceneManifest_Generation = YES;
 				INFOPLIST_KEY_UIApplicationSupportsIndirectInputEvents = YES;
 				INFOPLIST_KEY_UILaunchScreen_Generation = YES;
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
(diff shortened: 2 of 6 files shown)


kkostov:
@FahimF thanks for the contribution!
 
Renderers are usually something the consuming app should implement and/or customize, but we can definitely include a new platform in the ones that ship with TootSDK. 

I agree with your comment this change should be a new `AppKitAttribStringRenderer` and leave the UIKit renderer unchanged. For context, renderers are used as needed by the consuming app and one renderer should focus only on a single platform case.

I'm not sure why you had to make changes to the demo app, if you feel it can be improved, please open a separate PR.

Also, please register the new `AppKitAttribStringRenderer` renderer with the existing unit tests by appending the new platform target. (they should pass without changes)

FahimF:
@kkostov I might have started from the wrong place here 😄 I wanted to get the sample app to compile for macOS instead of being a Catalyst app. Everything else in TootSDK was fine except for the UIKitAttribStringRenderer. Hence the changes to both UIKitAttribStringRenderer and the demo app since the demo app was changed to compile for macOS.

Also, I tried to keep the same class (UIKitAttribStringRenderer) for the code so that the demo app did not need to have two separate class calls for each platform ... But perhaps that's not how you guys want to approach this?

Basically, I guess my question is whether you see UIKitAttribStringRenderer as a part of the library and if it should be cross-platform inherently, or if you want to have different renderers for each platform and want the developer to make the choice as to which renderer(s) to use for their application depending on the supported platforms?

kkostov:
@FahimF no worries, let me try to clarify. 

Renderers are indeed part of TootSDK and if you peek into [HTMLRendering](https://github.com/TootSDK/TootSDK/tree/main/Sources/TootSDK/HTMLRendering) you will see that we have several depending on the platform your want to use. These are provided out of the box to get you started and be able to show toots in your app right away.

Depending on your app's platform and setup, you will choose which TootSDK renderer to use:

```swift
// iOS, iPadOS, Catalyst
let renderer = UIKitAttribStringRenderer()

// macOS (AppKit)
let renderer = AppKitAttribStringRenderer()

// Server-side Swift, all platforms (very minimum featureset)
let renderer = UniversalRenderer()
```

Rendering toots natively depends on your use case and the kind of experience you want to build in your app. That's why you're also free to create your own renderers and use them to materialize toots as needed. In code, you'd normally instantiate a renderer and do something like `let content = renderer.render(post)`. You can see more usage examples in the [unit tests](https://github.com/TootSDK/TootSDK/blob/main/Tests/TootSDKTests/AttribStringRendererTests.swift).

If you find issues with any of the renderers above, you're welcome to submit a fix or an improvement 🤩!

Regarding the examples, so far we've opted to provide a separate example app for each platform. We have one for server side Swift, another for terminal/console applications and an example for iOS and Catalyst apps. We don't yet have an AppKit example so if you feel like one would be useful, you're more than welcome to create it!

FahimF:
@kkostov Ah, now it makes sense 😄 I had totally missed the fact that there was an existing `AppKitAttribStringRenderer` since I began from the sample app and thought only of extending it to support macOS natively.

I don't believe there is a need for a separate macOS example since you can easily add a macOS target to the existing sample app. I'll close this PR and submit another one later with the necessary change if that is something that you think would be useful?

davidgarywood:
I’d like that @FahimF. @kkostov and I are thinking about where next to take TootSDK for the next year or so at the moment. PRs like this are great in that they can show up areas we haven’t explained very easily or parts that we mentally filed under “todo” but haven’t got back around to. So thank you! 👍

FahimF:
@davidgarywood Sure thing 😄 

I've worked on 3 - 5 Mastodon clients at one time or another over the past year or two — Mastonaut, Tusky, and IceCubes are the ones I modified extensively and used for longer than a few weeks. In each case, I had to modify the API parsing to handle posts from my server since I use Akkoma and if I recall correctly, at least two of the above did not parse Akkoma responses correctly 😛 

I'd seen multiple references to TootSDK on Mastodon and had intended to replace the API client in my latest default Mastodon client iteration with TootSDK at some point but never found the time since it would have been a rather big job.

That's why I just took your sample app and started messing with it to compile it for macOS with the intention of creating yet another barebones Mastodon client to meet my current requirements. I must say that it works great and handles my Akkoma feeds without any issues at all 😄 

So, I totally appreciate what you guys have built here and would be happy to contribute what I can.

Will do another iteration of the macOS changes and submit a new PR over the next few days ...

PR #208, opened 2023-10-05
Native macOS Destination

FahimF:
* Remove the macOS Catalyst destination
* Add a new native macOS destination
* Change the few code instances that weren't compatible with macOS to allow compiling for macOS via precompiler directives

### DIFF
--- a/Examples/swiftui-toot/SwiftUI-Toot-Info.plist
+++ b/Examples/swiftui-toot/SwiftUI-Toot-Info.plist
@@ -15,12 +15,5 @@
 			</array>
 		</dict>
 	</array>
-	<key>UIApplicationSceneManifest</key>
-	<dict>
-		<key>UIApplicationSupportsMultipleScenes</key>
-		<true/>
-		<key>UISceneConfigurations</key>
-		<dict/>
-	</dict>
 </dict>
 </plist>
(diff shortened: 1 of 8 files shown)


kkostov:
Hi @FahimF, are you actively working on this PR?

Looking at the changes, I'm not exactly sure what are the "macOS specific" usage that we want to demonstrate. Or was it that you were fixing a bug when running on macOS (Designed for iPad)? 

Generally speaking, we prefer to keep a iOS/Catalyst example as is and without adding platform `#ifs` so it remains readable (we have a few, but we're actively removing them). If you see something we can showcase that's macOS specific, I'd opt for a dedicated macOS example instead.

FahimF:
@kkostov This is as per a discussion in a separate PR with relevance to this comment: https://github.com/TootSDK/TootSDK/pull/207#issuecomment-1746510334

Not everybody wants macOS support simply via Catalyst since the Catalyst approach has its own issues. So my effort was to simply show how easy it was to use TooSDK in a native macOS project and as per the above linked discussion. If that's not something you are interested in, please close the PR.

davidgarywood:
Yup, although, I wasn't expecting so much #ifdef code to have a macOS destination, back on that discussion! 

One of the things we've been thinking about is actually restructuring the examples away from one that's trying to do everything. We've been thinking that more detailed use-case scenarios might be better to have documented, using docc for laying them out. 

Appreciate the original effort with this PR, so thank you. I'll close it though, now.

Issue #211, opened 2023-10-16
TootClient.getAccountsBoosted(String) is missing pagination


f1xmAn:
**TootClient.getAccountsBoosted(String)** sends 
`GET /api/v1/statuses/:id/reblogged_by`
request w/o any pagination details. As a result, Mastodon server returns the same 40 accounts to each request.

[Mastodon API](https://docs.joinmastodon.org/methods/statuses/#reblogged_by) supports _max_id_ url request parameter to control pagination. It can be obtained from the response header _Link_.

Example of the Link response header:
```
Link: <https://mastodon.social/api/v1/statuses/111244178598174102/reblogged_by?max_id=111244833615323134>; rel="next", <https://mastodon.social/api/v1/statuses/111244178598174102/reblogged_by?since_id=111244934906844408>; rel="prev"
```

kkostov:
Thanks for spotting this @f1xmAn, a PR to address it is on the way!
Please note that pagination is only going to work for fedi flavours which support it. In case of Pleroma/Akkoma/Friendica - you will only get a single page of results.

PR #212, opened 2023-10-18
[BREAKING] Add pagination to favourited and boosted

kkostov:
Adresses #211 

* getAccountsBoosted is now paginated
* getAccountsFavourited is now paginated
* pagination only used for Masotdon, Pleroma-based flavours do not support pagination and so they will always return a single page


davidgarywood (approved):
Excellent - definitely an oversight from the initial dev !

PR #213, opened 2023-10-27
Add support for instance rules

daprice:
- Parse instance rules from instance info (https://docs.joinmastodon.org/entities/Instance/#rules) when using `TootClient.getInstanceInfo()`
- Ability to request just the rules from [`api/v1/instance/rules`](https://docs.joinmastodon.org/methods/instance/#rules) using `TootClient.getInstanceRules()`


Tunous (approved):

davidgarywood (approved):
Nice !

PR #214, opened 2023-10-31
Add support for reports

Tunous:
All flavours except for Pixelfed use:
- https://docs.joinmastodon.org/methods/reports/#post

Pixelfed doesn't implement that API but has its own. Unfortunately I didn't find public documentation so this implementation is based on source code.

- [Endpoint route](https://github.com/pixelfed/pixelfed/blob/4b9d0dc6efb1206ac51fc00e6f94c264da33d278/routes/api.php#L114)
- [Endpoint implementation](https://github.com/pixelfed/pixelfed/blob/4b9d0dc6efb1206ac51fc00e6f94c264da33d278/app/Http/Controllers/Api/ApiV1Dot1Controller.php#L68)
- Example from Vernissage (open source Pixelfed iOS app): [Reports.swift](https://github.com/VernissageApp/Vernissage/blob/2b88870995366de53ac99b250f68e2f928e5482b/PixelfedKit/Sources/PixelfedKit/Targets/Reports.swift#L15), [Report.swift](https://github.com/VernissageApp/Vernissage/blob/2b88870995366de53ac99b250f68e2f928e5482b/PixelfedKit/Sources/PixelfedKit/Entities/Report.swift#L10C32-L10C32)

---

In implementation I decided to differentiate these by a different type of params. Alternatively could rename one of the functions. Didn't see a way to use single function as the calls are too different.

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Reports.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Reports.swift
@@ -70,7 +70,7 @@ extension TootClient {
         if ReportCategory.mastodonSupported.contains(params.category) {
             queryItems.append(.init(name: "category", value: params.category.rawValue))
         }
-        for ruleId in params.ruleIds ?? [] {
+        for ruleId in params.ruleIds {
             queryItems.append(.init(name: "rule_ids[]", value: String(ruleId)))
         }
         return queryItems
--- a/Sources/TootSDK/TootClient/TootClient+Reports.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Reports.swift
@@ -1,0 +1,63 @@
+//
+//  TootClient+Reports.swift
+//  
+//
+//  Created by Łukasz Rutkowski on 26/10/2023.
+//
+
+import Foundation
+
+public extension TootClient {
+
+    /// Report problematic users or posts to moderators.
+    ///
+    /// - Warning: For Pixelfed use `PixelfedReportParams`.
+    @discardableResult
+    func report(_ params: ReportParams) async throws -> Report {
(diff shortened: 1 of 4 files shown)


Sources/TootSDK/TootClient/TootClient+Reports.swift
```diff
@@ -0,0 +1,63 @@
...
+
+import Foundation
+
+public extension TootClient {
+
+    /// Report problematic users or posts to moderators.
+    ///
+    /// - Warning: For Pixelfed use `PixelfedReportParams`.
+    @discardableResult
+    func report(_ params: ReportParams) async throws -> Report {
```
kkostov, line 16: I think we need to find a way to use the same signature for requesting the report from all flavours. Change in behaviour based on flavour should be transparent for the client. We have examples of such methods `followAccountURI` and `updateCredentials`.

In this case, the params object could be a higher-level API provided by TootSDK which (depending on flavour) we will use internally to create the appropriate server request.

Tunous:
Unified the calls to use single params struct. In order for this to work properly I made following changes.

Category is now a required property. Additionally to allow clients to determine which categories are valid I changed the category parameter to an enum with all known values. Clients can access new `reportCategories` in `TootClient` to determine which options to offer to users.

Pixelfed will throw error if unsupported category is used, other flavours will ignore it, essentially marking the report as "other".
For pixelfed if `postIds` is provided the first id will be used and report type will be set to "post". Otherwise report will be set to type "user" with id from `accountId` property.
All other parameters are ignored.

Tunous:
One small change to not parse responses because Friendica returns empty arrays.

kkostov (approved):
Nice! Thank you so much

PR #215, opened 2023-11-05
Add public init to Poll.Option

boscojwho:
Why:
- Allow SDK users to create mock data for use with SwiftUI #Preview.

Notes:
- Currently, this would involve decoding a mock JSON payload containing Poll data, which is considerably more work than writing some mock data in Swift.


davidgarywood:
Hi @boscojwho  - do you want to stand up a quick PR with this in ?

davidgarywood (approved):

kkostov (approved):
Awesome!

PR #216, opened 2023-11-20
support for reading announcements and marking an announcement as read

technicat:
https://docs.joinmastodon.org/methods/announcements/

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Announcements.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Announcements.swift
@@ -1,14 +1,14 @@
 //
 //  TootClient+Announcements.swift
-//  
+//
 //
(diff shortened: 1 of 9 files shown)


Sources/TootSDK/TootClient/TootClient+Announcements.swift
```diff
@@ -0,0 +1,41 @@
...
+
+        _ = try await fetch(req: req)
+    }
+}
+
+extension TootFeature {
+
+    /// Ability to retrieve announcements
+    ///
+    public static let announcements = TootFeature(supportedFlavours: [.mastodon, .akkoma, .pleroma])
```
kkostov, line 40: PixelFed doesn't support this?

kkostov (approved):
Looks great! And thanks for linting the project 🤩!

PR #219, opened 2023-12-02
Add support for markers API

Tunous:
Marked as not available for:
- firefish: returns 404.
- pixelfed: both endpoints always return empty array for me which doesn't decode.


davidgarywood (approved):
Thanks for this ! 👍

PR #220, opened 2023-12-03
add support for recognizing a post that should be filtered (i.e. filtered with warning)

technicat:
https://docs.joinmastodon.org/entities/Status/#filtered

### DIFF
--- a/Sources/TootSDK/Models/Filter.swift
+++ b/Sources/TootSDK/Models/Filter.swift
@@ -18,7 +18,7 @@ public struct Filter: Codable, Hashable, Identifiable {
     }
 
     public enum Action: String, Codable {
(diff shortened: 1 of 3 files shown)


technicat:
> @technicat this is great! 🎊 Could we rename mentions of "status" to "post" in the library API and comments?

You mean change FilterStatus to FilterPost and statusMatches to postMatches? Probably should do the same with Announcement.statuses and Announcement.Status, too.

technicat:
> @technicat indeed, but that would be a breaking change at this point (something we can do in v2) - let's limit the change to renaming the newly added fields, structs and comments.

Announcement.statuses was only added yesterday.

kkostov:
@technicat just the ones in this PR, I logged an issue to tackle the others in a separate PR.

kkostov (approved):

PR #223, opened 2023-12-12
fix: handle Friendica returning [] instead of nil for contact account

technicat:
Added a decoder for Instance, handles the case of Friendica (example cambrian.social) returning [] instead of nil for no contact account, resulting in this error

```
[TootSDK bug] There was an error decoding incoming data:
Type 'Dictionary<String, Any>' mismatch:Expected to decode Dictionary<String, Any> but found an array instead.
codingPath:[CodingKeys(stringValue: "contactAccount", intValue: nil)].
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -52,6 +52,7 @@ let package = Package(
                 .copy("Resources/familiar_followers_nofollowers.json"),
                 .copy("Resources/featured_tag.json"),
                 .copy("Resources/featured_tag_unused.json"),
+                .copy("Resources/instance_friendica_noaccount.json"),
                 .copy("Resources/list.json"),
                 .copy("Resources/post no emojis.json"),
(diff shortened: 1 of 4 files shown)


kkostov:
@technicat would it be possible to add a test for this case? If you have a friendica response to use as a test data that would be great 🤩

technicat:
> @technicat would it be possible to add a test for this case? If you have a friendica response to use as a test data that would be great 🤩

Most of the cases I see have an email address in the data, which makes sense since there's no account contact, but I did fine one that has neither.

kkostov:
@technicat Good thinking, we can also replace them with "social.thisworksonmymachine.local" or something to avoid accidentally referring something real.

kkostov (approved):

PR #224, opened 2023-12-12
Add support for follow requests

Tunous:


Examples/swiftyadmin/Sources/swiftyadmin/FollowRequests/AcceptFollowRequest.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+import ArgumentParser
+import TootSDK
+
+struct AcceptFollowRequest: AsyncParsableCommand {
+    @OptionGroup var auth: AuthOptions
+    @Option var id: String
+
+    mutating func run() async throws {
+        let client = try await auth.connect()
```
kkostov, line 17: There is no need to call `connect` if the async init `try await TootClient(connect: url, accessToken: token)` is used as it calls connect internally.

I like the AuthOptions refactor, but I'd leave the bit that creates a `TootClient` instance to remain in every example so that one coming here to check how things are working won't have to browse the code (or know how `OptionGroup` works).

Tunous: Done

kkostov (approved):

PR #232, opened 2023-12-21
feat: added TootClient.getExtendedDescription()

technicat:
https://docs.joinmastodon.org/methods/instance/#extended_description


kkostov (approved):
Awesome! Thank you @technicat 🚀!

davidgarywood:
Ah, just the way I like it 😅

PR #233, opened 2023-12-27
flavour check for sharkey and iceshrimp

technicat:
For now, recognize iceshrimp as firefish, same as catodon (both firefish forks). Sharkey is a misskey fork but separate from firefish.

Split the trends feature into separate features for trending posts/tags/links. Firefish only supports trending posts, and sharkey only supports trending tags.

### DIFF
--- a/Sources/TootSDK/Models/Instance.swift
+++ b/Sources/TootSDK/Models/Instance.swift
@@ -207,21 +207,28 @@ extension Instance {
 
 extension Instance {
     public var flavour: TootSDKFlavour {
+        // 2.7.2 (compatible; Pleroma 2.5.0)
         if version.lowercased().contains("pleroma") {
             return .pleroma
         }
+        // 2.7.2 (compatible; Pixelfed 0.11.9)
         if version.lowercased().contains("pixelfed") {
             return .pixelfed
         }
+        // 2.8.0 (compatible; Friendica 2023.05)
         if version.lowercased().contains("friendica") {
             return .friendica
         }
+        // 2.7.2 (compatible; Akkoma 3.10.4-0-gebfb617)
         if version.lowercased().contains("akkoma") {
             return .akkoma
         }
+        // 3.0.0 (compatible; Firefish 1.0.4-dev5)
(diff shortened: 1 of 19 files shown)


Sources/TootSDK/Models/Instance.swift
```diff
@@ -219,9 +219,12 @@ extension Instance {
         if version.lowercased().contains("akkoma") {
             return .akkoma
         }
-        if version.lowercased().contains("firefish") {
+        if version.lowercased().contains("firefish") || version.lowercased().contains("iceshrimp") {
```
kkostov, line 222: With the additional logics, would it be possible to add some unit tests around the new flavour detections?

Sources/TootSDK/TootClient/TootClient+Trends.swift
```diff
@@ -33,7 +33,7 @@ extension TootClient {
     ///   - offset: Skip the first n results.
     /// - Returns: Array of ``Post``.
     public func getTrendingPosts(limit: Int? = nil, offset: Int? = nil) async throws -> [Post] {
-        try requireFeature(.trends)
```
kkostov, line 36: Maybe I'm not seeing it, why does the feature need to be renamed to support the new flavour?

kkostov:
@technicat Love it, thank you for adding this. Would it be possible to add a couple of tests around the flavour detection so that we have concrete version string examples in the repo?

technicat:
> @technicat Love it, thank you for adding this. Would it be possible to add a couple of tests around the flavour detection so that we have concrete version string examples in the repo?

Do you want to anonymize the domains in the same way as the existing friendica test? They also have real contact info: name, uri, and email. What do you want in place of that?

kkostov:
> Do you want to anonymize the domains in the same way as the existing friendica test? They also have real contact info: name, uri, and email. What do you want in place of that?

Indeed, if committing new samples, please remove or replace identifying information with lorem ipsum or similar.

kkostov (approved):
🚀

PR #235, opened 2023-12-30
Web Push message decrypting API

Tunous:
This is a WIP pull request about including Web Push message handling code in TootSDK. Right now I've included a method to decrypt message received from a push.

Some questions:
- Decoding uses CryptoKit. I have no idea how this works on Linux so I've wrapped it in canImport check. Is that enough or is there an alternative?
- Additionally included conversion from URL safe Base 64 which is the format in which all components are sent. Again I have no idea how this works on Linux but the used methods are from Foundation.

TODO:
- `PushNotification` decoding. Should decrypt method be private and instead only make public one which will decrypt and convert to this model instead?
- Version of the method which will accept parameters in their base64url encoded form?
- Extension on push subscription model which will accept keys using models from CryptoKit for consistency
- Extension to generate valid auth secret


kkostov:
Regarding CryptoKit, it will not be compatible with linux targets (e.g. server-side management of notifications), so I suggest that we use https://github.com/apple/swift-crypto instead.

kkostov:
> Additionally included conversion from URL safe Base 64 which is the format in which all components are sent. Again I have no idea how this works on Linux but the used methods are from Foundation.

I'm just wondering here, do we have to convert the base64 string? For example, when making a `post` to `/api/v1/push/subscription`, we can use form encoding for the body:

```swift
let req = try HTTPRequestBuilder {
            $0.url = getURL(["api", "v1", "push", "subscription"])
            $0.method = .post
            $0.body = try .form(queryItems: createQuery(from: params))
        }
```

Then we can use `Data.base64EncodedString()` and `Data(base64Encoded: String)`.

kkostov:
> PushNotification decoding. Should decrypt method be private and instead only make public one which will decrypt and convert to this model instead?

I think we need some examples of payloads. It's possible we can't provide a single model that can handle the payload from each flavour and we'd need this to be behind a flavour check. This means that when a notification is received by the app, it will need to pass it to a connected `TootClient` instance to have it parsed.

> Version of the method which will accept parameters in their base64url encoded form?

Indeed, I suspect the public API method will have to receive the raw base64 string for the reason mentioned above - we don't have assurance it's always the same model I think.

> Extension to generate valid auth secret

Sure, with swift-crypto, it can look something like this (pseudo-code):

```swift
final class PushSubscription {
  init(encodedKey: String) {
    self.encodedKey = encodedKey
  }

  var encodedKey: String

  var privateKey: P256.KeyAgreement.PrivateKey {
    let data = Data(base64Encoded: self.encodedKey)!
    return try! P256.KeyAgreement.PrivateKey(rawRepresentation: data)
  }

  var publicKey: P256.KeyAgreement.PublicKey {
    return self.privateKey.publicKey
  }

  var p256dh: String {
    return self.publicKey.x963Representation.base64EncodedString()
  }

  init() {}
}
```

and some helpers to make it:

```swift
extension PushSubscription {
  static func new() -> PushSubscription {
    // create a new key
    let key = P256.KeyAgreement.PrivateKey()
    let encodedKey = key.rawRepresentation.base64EncodedString()
    return PushSubscription(encodedKey: encodedKey)
  }
}
```

Tunous:
> I'm just wondering here, do we have to convert the base64 string? For example, when making a `post` to `/api/v1/push/subscription`, we can use form encoding for the body:

Yes, we have to convert them. When sent as regular base64 string, subscription call succeeds but mastodon never calls the proxy endpoint. ` .form(queryItems:)` doesn't appear to be doing necessary conversions and I'm not sure whether it should be modified.

> I think we need some examples of payloads. It's possible we can't provide a single model that can handle the payload from each flavour and we'd need this to be behind a flavour check. This means that when a notification is received by the app, it will need to pass it to a connected `TootClient` instance to have it parsed.

Mastodon and Friendica return json of `PushNotification` I added in this PR. Not sure about pleroma and akkoma yet as apparently the instance I'm using is defederated. Need to find a new server for test account. Any recommendations?

kkostov:
> Need to find a new server for test account. Any recommendations?

I haven't had a chance to use an Akkoma recently. I guess any one would do? If it's hard to find a decent one, I could setup a dedicated akkoma instance for testing.

Tunous:
Yeah, Akkoma is a bit problematic. I got test account on instance https://miraiverse.xyz and it seems to call the provided endpoint differently than other flavours causing my proxy to fail. But that's unrelated to TootSDK.

Pleroma confirmed to use the same payload as Mastodon with extra pleroma field that I'm ignoring. Documentation/source for akkoma indicates the same.

Marking as ready for review as I see no further issues in added code.

kkostov:
Awesome @Tunous ! Thank you so much. Code-wise it looks good.

I will update the README to include a few words on how to use this. Are you using an open-source proxy for the notifications? If yes, would you mind sharing it, so I can check it out? I would like to have a full "end to end" recipe for notifications in the README (for folks to get started quickly before diving into advanced scenarios).

Tunous:
> Are you using an open-source proxy for the notifications?

Yes https://github.com/metabolist/metatext-apns which is mentioned in Mastodon documentation.

kkostov (approved):

PR #237, opened 2024-01-06
Notifications API improvements

Tunous:
This pull request introduces lots of changes related to notifications and push notifications API. Needed to correctly implement notifications support across all available flavours in SDK.

- Added option to configure `update` notification alert in init
- Added optional `policy` to push subscription response
- Changed `policy` to enum with known values
- Added `static NotificationType.supported(by:)`, `NotificationType.isSupported(by:)` to determine if given flavour supports selected notification type
- Added `static NotificationType.supportedAsPush(by:)`, `NotificationType.isSupportedAsPush(by:)` to determine if given flavour supports selected notification type in push subscriptions
- Added mapping from `types` to `excludeTypes` for Friendica which doesn't support the former
- Fixed akkoma and pleroma not respecting `types` parameter
- Added `TootFeature.pushSubscriptionsPolicy` to determine if push subscriptions support policy setting
- [BREAKING] Changed `types` and `excludeTypes` in `TootNotificationParams` to Set
- Fixed update subscription call was missing `repost` query item

The supported notification types and policy were verified manually via trial and error plus source code analysis due to lack of documentation. I'm not 100% sure whether it's all correct.

---

Other issues I identified but was unable to resolve on TootSDK side:
- Friendica (venera.social) returns duplicate push subscription json from call to POST /v1/push/subscription causing response parsing to fail
- Friendica (venera.social) returns 403 instead of 404 from call to GET /v1/push/subscription if there is no subscription
- Friendica (venera.social) returns 403 from successful call to DELETE /v1/push/subscription
- Friendica (venera.social) does not update subscription after call to PUT /v1/push/subscription

### DIFF
--- a/Sources/TootSDK/Models/PushSubscription.swift
+++ b/Sources/TootSDK/Models/PushSubscription.swift
@@ -5,10 +5,11 @@ import Foundation
 
 /// Represents a subscription to the push streaming server.
 public struct PushSubscription: Codable, Sendable {
-    public init(endpoint: String, alerts: Alerts, serverKey: String) {
+    public init(endpoint: String, alerts: Alerts, serverKey: String, policy: PushSubscriptionPolicy?) {
         self.endpoint = endpoint
         self.alerts = alerts
         self.serverKey = serverKey
+        self.policy = policy
     }
 
     /// Where push alerts will be sent to.
@@ -17,4 +18,6 @@ public struct PushSubscription: Codable, Sendable {
     public var alerts: Alerts
     /// The streaming server's VAPID key.
     public var serverKey: String
+    /// From who to receive push notifications.
+    public var policy: PushSubscriptionPolicy?
 }
--- a/Sources/TootSDK/Models/PushSubscription.swift
+++ b/Sources/TootSDK/Models/PushSubscription.swift
@@ -5,7 +5,7 @@ import Foundation
 
 /// Represents a subscription to the push streaming server.
 public struct PushSubscription: Codable, Sendable {
-    public init(endpoint: String, alerts: Alerts, serverKey: String, policy: PushSubscriptionPolicy?) {
+    public init(endpoint: String, alerts: Alerts, serverKey: String, policy: PushSubscriptionPolicy = .all) {
         self.endpoint = endpoint
         self.alerts = alerts
         self.serverKey = serverKey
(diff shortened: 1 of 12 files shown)


Sources/TootSDK/Models/PushSubscription.swift
```diff
@@ -5,10 +5,11 @@ import Foundation
 
 /// Represents a subscription to the push streaming server.
 public struct PushSubscription: Codable, Sendable {
-    public init(endpoint: String, alerts: Alerts, serverKey: String) {
+    public init(endpoint: String, alerts: Alerts, serverKey: String, policy: PushSubscriptionPolicy?) {
```
kkostov, line 8: Can we give policy a default value in the init, for example `.all`?

kkostov, line 8:
```suggestion
    public init(endpoint: String, alerts: Alerts, serverKey: String, policy: PushSubscriptionPolicy = .all) {
```

Sources/TootSDK/Models/TootNotificationParams.swift
```diff
@@ -19,12 +22,28 @@ public struct TootNotificationParams: Codable, Sendable {
...
+    public var types: Set<TootNotification.NotificationType>?
 
     enum CodingKeys: String, CodingKey {
         case excludeTypes = "exclude_types"
         case types = "types"
     }
 }
+
+extension TootNotificationParams {
+    func corrected(for flavour: TootSDKFlavour) -> TootNotificationParams {
```
kkostov, line 36:
```suggestion
    private func adapted(for flavour: TootSDKFlavour) -> TootNotificationParams {
```

kkostov:
This is fantastic work @Tunous Thank you 🙏!
Regarding Friendica, some of these sound like actual bugs on their server side. I opened a separate issue to keep track (as they deserve dedicated PRs and probably some testing).

Sources/TootSDK/Models/TootNotificationParams.swift
Tunous: It is inheriting its default internal visibility which is needed to call it from other files in module. I can add explicit `internal` if you prefer.

Tunous:
@kkostov Addressed issues and updated for sharkey based on source code (not verified via api calls)

kkostov (approved):

PR #241, opened 2024-01-11
Pixelfed and Sharkey fixes

technicat:
For pixelfed, adjust the Account decoder to handle missing header and emojis fields, which happens when querying the blocked and muted account lists.

Sharkey doesn't support muting and bookmarking posts and dismissing single notifications, so added feature checks for those.

Sharkey also generates errors for requests with empty bodies but the content-type set to application/json, which could be considered incorrect as discussed in the links below. At first I started to add dummy JSON content for the bodies but it seems more correct to omit the content-type in those cases and specify content-length: 0

https://stackoverflow.com/questions/40614416/is-empty-body-correct-if-content-type-is-application-json#40624629
https://stackoverflow.com/questions/29784398/should-content-type-header-be-present-when-the-message-body-is-empty
https://stackoverflow.com/questions/4191593/is-it-considered-bad-practice-to-perform-http-post-without-entity-body

```json
{
"code" : "FST_ERR_CTP_EMPTY_JSON_BODY",
"statusCode" : 400,
"message" : "Body cannot be empty when content-type is set to 'application/json'",
"error" : "Bad Request"
}
```


Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -160,10 +160,16 @@ extension TootClient {
 
     /// Fetch data asynchronously and return the raw response.
     internal func fetch(req: HTTPRequestBuilder) async throws -> (Data, HTTPURLResponse) {
-        if req.headers.index(forKey: "Content-Type") == nil {
+        if req.headers.index(forKey: "Content-Type") == nil && req.body != nil {
```
kkostov, line 163: I would prefer if the changes to the request are specific to the flavour which mandates them. In this case, if flavour is sharkey, then we apply the content type modifications.

Not all flavours handle this in the same way. For example, Pleroma/Akkoma would always look for the content-type header and default to "application/json" if it's not set. So the behaviour and compatibility may vary. It will be best to make such instances explicit.

Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -160,14 +160,15 @@ extension TootClient {
 
     /// Fetch data asynchronously and return the raw response.
     internal func fetch(req: HTTPRequestBuilder) async throws -> (Data, HTTPURLResponse) {
-        if req.headers.index(forKey: "Content-Type") == nil && req.body != nil {
+        if req.headers.index(forKey: "Content-Type") == nil {
```
kkostov, line 163: I think this check is also sharkey specific.

technicat: I undid my change and moved it to a separate if clause that checks for sharkey (although I don't consider it sharkey-specific, seems to me more generally correct and works with all platforms I tested but happens to be necessary for sharkey). The files changed shows that first part is the same as before.

PR #242, opened 2024-01-14
Add follow-account example command

Tunous:
With this change it will be possible to update account following parameters one by one. For example to be able to enable account notifications without changing boosts following status.
Default values are set to `nil` to follow server preferences.

### DIFF
--- a/Examples/swiftyadmin/Package.resolved
+++ b/Examples/swiftyadmin/Package.resolved
@@ -28,6 +28,15 @@
       }
     },
     {
+      "identity" : "swift-crypto",
+      "kind" : "remoteSourceControl",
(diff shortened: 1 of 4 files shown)


davidgarywood:
Hi @Tunous - this seems like a good idea.

 I can see these aren’t required fields on the Mastodon API, so this should work fine, have you confirmed any of the other instance types at all though?

Tunous:
@davidgarywood Good point, of course I forgot to check that. After verification with pleroma I see that it always modifies all parameters. If I send only `notify` it also changes `reposts`.
I see no way to resolve this correctly with optionals. Based on that I reverted my changes. All that's left is just a new example command.

kkostov (approved):
Thank you @Tunous !

PR #244, opened 2024-01-28
add default empty string param for ExtendedDescription

technicat:
so you can instantiate with

ExtendedDescripion()

vs

ExtendedDescription(content: "")

edit: catodon.social updated today with code newly synced to iceshrimp but with catodon added to the version string, so updated the TootSDK flavour check accordingly


technicat:
> Thanks, @technicat! Btw, it will be nice to have different changes in separate PRs, so they can be reviewed, tested and listed individually. Even if the changes are small.

Yeah I considered removing the first change, the catodon fix is more urgent, since their update breaks the previous catodon compatibility.

PR #247, opened 2024-02-07
Make order of attachments for ScheduledPost consistent

### LINKED ISSUE #246: Inconsistent order of media attachments for scheduled posts
Hey 👋,
I'm building an app to schedule posts and noticed that the `mediaAttachments` and `params.mediaIds` in `ScheduledPost` have inconsistent ordering. `params.mediaIds` appears to have the correct order, while `mediaAttachments` are sorted numerically by their id. I'm not sure whether this would be considered to be a bug in the Mastodon API, but it would be nice to have consistent ordering in TootSDK. If you agree, I can create a pull request to add a decodable implementation of `ScheduledPost` that fixes the ordering based on `params.mediaIds`.

mikakruschel:
This fixes #246

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -74,6 +74,7 @@ let package = Package(
                 .copy("Resources/relationship.json"),
                 .copy("Resources/scheduled_post_sensitive.json"),
                 .copy("Resources/scheduled_post_attachment.json"),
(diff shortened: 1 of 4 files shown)


kkostov:
This is great, thank you @mikakruschel 🚀!

kkostov (approved):

PR #252, opened 2024-02-12
support for Mastodon translation functions

### LINKED ISSUE #245: Post.application init param doesn't match the property
The Post application property is optional so the corresponding init param should be changed from

```swift
application: TootApplication,
```
to
```swift
application: TootApplication?,
```

technicat:
Access to the translation functions added in Mastodon 4.x via TootClient.getPostTranslation and TootClient.getTranslationLanguages. Includes fix #245 to facilitate Post copy (I place the translation results in a copy of the original post, the Post copy init is in a TootSDK extension package at https://github.com/technicat/fedicat)

### DIFF
--- a/Sources/TootSDK/Models/PostTranslationParams.swift
+++ b/Sources/TootSDK/Models/PostTranslationParams.swift
@@ -1,6 +1,6 @@
 //
 //  TranslatePostParams.swift
-//  
+//
 //
 //  Created by Philip Chu on 1/30/24.
 //
--- a/Sources/TootSDK/Models/PostTranslationParams.swift
+++ b/Sources/TootSDK/Models/PostTranslationParams.swift
@@ -10,7 +10,7 @@ import Foundation
 /// Params to translate a post
 public struct PostTranslationParams: Codable {
     /// (ISO 639 language code).
-    /// The status content will be translated into this language.
+    /// The post content will be translated into this language.
     /// Defaults to the user’s current locale.
     public var lang: String
 
--- a/Sources/TootSDK/Models/PostTranslationParams.swift
+++ b/Sources/TootSDK/Models/PostTranslationParams.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 /// Params to translate a post
-public struct TranslatePostParams: Codable {
+public struct PostTranslationParams: Codable {
     /// (ISO 639 language code).
     /// The status content will be translated into this language.
     /// Defaults to the user’s current locale.
--- a/Sources/TootSDK/Models/Translation.swift
+++ b/Sources/TootSDK/Models/Translation.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 public struct Translation: Codable {
-    
+
     /// HTML-encoded translated content of the status.
     public var content: String
     /// The translated spoiler warning of the status.
@@ -27,9 +27,9 @@ public struct Translation: Codable {
     /// The target language
     /// Not in the Mastodon spec, but in the Mastodon code
     public var language: String
-    
+
     public struct MediaAttachment: Codable, Hashable, Identifiable, Sendable {
-        
+
         public init(
             id: String,
             description: String? = nil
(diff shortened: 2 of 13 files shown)


Sources/TootSDK/Models/PostTranslationParams.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Philip Chu on 1/30/24.
+//
+
+import Foundation
+
+/// Params to translate a post
+public struct PostTranslationParams: Codable {
+    /// (ISO 639 language code).
+    /// The status content will be translated into this language.
```
kkostov, line 13:
```suggestion
    /// The post content will be translated into this language.
```

Sources/TootSDK/Models/Translation.swift
```diff
@@ -0,0 +1,70 @@
...
+    /// The language of the source text, as auto-detected by the machine translation provider.
+    /// ISO 639 language code
+    public var detectedSourceLanguage: String
+    /// The service that provided the machine translation.
+    public var provider: String
+    /// The target language
+    /// Not in the Mastodon spec, but in the Mastodon code
+    public var language: String
+
+    public struct MediaAttachment: Codable, Hashable, Identifiable, Sendable {
```
kkostov, line 31: I'd like the name here to better reflect that this is not an actual attachment but a translation for an existing media.

```suggestion
    public struct MediaAttachmentTranslation: Codable, Hashable, Identifiable, Sendable {
```

Sources/TootSDK/Models/Translation.swift
```diff
@@ -0,0 +1,70 @@
...
+            self.description = description
+        }
+
+        /// The ID of the attachment in the database.
+        public var id: String
+        /// Alternate text that describes what is in the media attachment, to be used for the visually impaired or when media attachments do not load.
+        public var description: String?
+    }
+
+    public struct Poll: Codable, Hashable, Identifiable, Sendable {
```
kkostov, line 47:
```suggestion
    public struct PollTranslation: Codable, Hashable, Identifiable, Sendable {
```

Sources/TootSDK/TootClient/TootClient+Post.swift
```diff
@@ -242,6 +242,20 @@ extension TootClient {
 
 }
 
+extension TootClient {
+
+    /// Translate the post content into some language.
+    public func getPostTranslation(id: String, params: PostTranslationParams? = nil) async throws -> Translation {
```
kkostov, line 248:
```suggestion
    public fund publishPostTranslation(id: String, params: PostTranslationParams? = nil) async throws -> Translation {
```

technicat: I can commit this but I believe it's wrong, you're not publishing a translation you're querying a translation.

Sources/TootSDK/Models/Translation.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  Translation.swift
+//
+//
+//  Created by Philip Chu on 1/30/24.
+//
+
+import Foundation
+
+public struct Translation: Codable {
```
kkostov, line 10: I think none of the translation attributes are guaranteed to be provided. The fields should be optional.

Sources/TootSDK/TootClient/TootClient+Post.swift
kkostov: Oh you're right of course!

kkostov:
Thanks a lot @technicat. I noted a few minor tweaks, let me know if you will get a chance to apply them or I could do so on my side. Also thanks for including a fix for #245 ! 🎉

technicat:
I'll hold off on more changes (don't agree with the get->publish renaming, PollTranslation makes more sense after it's detached otherwise we have Translation.PollTranslation, and more optionals might be a good idea in Translation but for now it matches the spec (which actuallly is pretty buggy) and all the test cases I've tried and so far Mastodon is the only implementation. You can change it how you like, of course.

kkostov:
Awesome @technicat, looks perfect to me :) 
I do agree with your comments for the naming - all good suggestions.

kkostov (approved):

PR #269, opened 2024-03-25
Support Mastodon's `legal` category for reporting illegal content

daprice:
This report category isn’t in Mastodon’s docs yet, but was added in mastodon/mastodon#23941 and released in Mastodon 4.2.0.

### DIFF
--- a/Sources/TootSDK/Models/ReportCategory.swift
+++ b/Sources/TootSDK/Models/ReportCategory.swift
@@ -9,6 +9,8 @@ import Foundation
 
 public enum ReportCategory: String, CaseIterable, Codable {
     case spam
+    /// Content that is illegal in the user's or the instance's country.
+    case legal
     case sensitive
     case abusive
     case underage
@@ -34,6 +36,7 @@ public enum ReportCategory: String, CaseIterable, Codable {
 
     public static let mastodonSupported: Set<ReportCategory> = [
         .spam,
+        .legal,
         .other,
         .violation,
     ]


Tunous:
Do you know what will happen if we send it for mastodon version < 4.2.0? Because if this doesn't work we'll need a way to know which version is used by instance.

kkostov (approved):
I don't see a way to reliably detect the availability of the `legal` option through the API (apart from a version check). We can get the list of custom rules, but that's about it.

Generally, we haven't been adding version checks to flavours e.g. we support Mastodon 4.x features and make them available without verifying if the instance is not running 3.8 or older. So in that sense, I'm ok with extending the enumeration for the entire mastodon flavour.

Alternatively, we could introduce additional flavours for major versions of specific server types e.g. `mastodon3`, `mastodon42` etc but that deserves a PR on itself as it will be a deeper change.

technicat:
I see although legal isn't documented it is mentioned in the version history

https://docs.joinmastodon.org/methods/reports/

I have class-based feature checking in my fedicat package that includes version checking so I can make sure the right stuff shows up or doesn't show up in the UI (I would actually prefer generally that not be enforced at the TootSDK level). Preparing for this change I have this in a MastoAPI class (and an override in a Pixelfed class with its own different set).

```swift
open var reportCats: [ReportCategory] {
    var cats: [ReportCategory] = [
      .spam,
      .other,
      .violation,
    ]
    if version >= Version(4, 2) {
    //  cats.append(.legal)
      return cats
    }
    return cats
  }
```

davidgarywood:
I agree with @kkostov here, happy to approve this in that we’re supporting 4x without considering any differences for older 👍

Also, love @technicat’s approach there too at the client side.

Version checking in TootSDK beyond flavour checking definitely needs to be discussed separately and come as its own piece of work.

(Approving!)

davidgarywood (approved):

Issue #268, opened 2024-03-23
snake_case for EditPostParams


technicat:
`EditPostParams` doesn't have snake_case mappings in its coding keys as in `PostParams` and `ScheduledPostRequest`,  and for example a call to `TootClient.editPost` will always remove any existing attachments since it issues `mediaIds[]` in the form instead of `media_ids[]` as in the spec

https://docs.joinmastodon.org/methods/statuses/#edit

So snake_case should be listed for mediaIds and also spoilerText (and also mediaAttributes/media_attributes which hasn't been added yet) like this

```swift
enum CodingKeys: String, CodingKey {
        case post = "status"
        case spoilerText = "spoiler_text"
        case sensitive
        case mediaIds = "media_ids"
        case poll
    }
```

PR #270, opened 2024-03-27
Updated Account to latest fields, supporting more account preferences, updated Post edit to include languge

technicat:
Added noindex, indexable, hideCollections, and role to Account, which supports opting in/out of search and hiding follower/followee info from the  public, and updated/added TootSource and TootRole accordingly (see added test cases for reference)

Also fixed missing snake case mapping in PostEditParams which was losing spoilers and attachments.


technicat:
fixes #268

kkostov:
@technicat That's quite the change! Thank you so much 🙌

✅ `indexable`, `noindex` and `hideCollections` looks good.

✅ `EditPostParams` looks good.

✅ `TootSource` looks good.

✅ `UpdateCredentialsParams` looks good.

✅ 🎉 Unit Tests ❤️


Regarding the `role` field, I think we need a little bit more to support it. Can we take it out of this PR?

Some thoughts: `Pleroma` uses "roles" as an object with a boolean field for "admin" or "moderator" which we can handle during serialization. I like the `TootRole` concept, perhaps we could simplify this object a bit to allow us to fill-in values when coming from other flavours (e.g. remove the `id` field, and change `color`, `permissions` and `highlighted` to Optional) Also, Mastodon only returns a role for "CredentialAccount", not regular "Account" so we should consider creating a dedicated model for the endpoints which actually return a role field.

kkostov:
@technicat On second thought, I'm OK to merge this if we can simplify `TootRule` so tweaking the serialization for pleroma will not cause too much breakage.

kkostov:
@technicat On second, second thought 😅, it seems supporting the Admin API calls for Pleroma/Akkoma's roles will be very much a dedicated effort as they require an entirely specific set of endpoints and models. Let's go with the Mastodon's role for now.

kkostov (approved):

davidgarywood (approved):

PR #271, opened 2024-03-28
Add support for `hint` property on `InstanceRule`

daprice:
This is another thing that was recently added to Mastodon (mastodon/mastodon#29539) but not officially released or documented yet (although mastodon.social and mastodon.online already have it deployed).

This one shouldn’t have any impact on usage with servers that don’t support it, because it’s optional.


kkostov (approved):
🚀

davidgarywood (approved):

PR #275, opened 2024-04-17
added optional withSuspended param to TootClient.getRelationships (specifies whether to ignore suspended accounts)

technicat:
Introduced in Mastodon 4.3.0

https://docs.joinmastodon.org/methods/accounts/#relationships

Passed via a RelationshipParams struct in the same manner as params in getAnnouncements (and others).


kkostov (approved):
✨

davidgarywood (approved):
🙏

PR #277, opened 2024-04-21
add Post.editedAt ("edited_at")

technicat:
Added relative to other Post properties same as in the API doc:

https://docs.joinmastodon.org/entities/Status/#edited_at


Tests/TootSDKTests/PostTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        XCTAssertEqual(result.id, "112300188527054704")
+        XCTAssertEqual(result.filtered, [])
+        XCTAssertEqual(result.mentions, [])
+        XCTAssertEqual(result.emojis, [])
+        XCTAssertEqual(result.mediaAttachments, [])
+        XCTAssertEqual(result.tags.count, 1)
+        XCTAssertEqual(result.favourited, false)
+        XCTAssertEqual(result.reposted, false)
+        XCTAssertNotNil(result.createdAt)
+        XCTAssertNotNil(result.editedAt)
```
kkostov, line 29: Could you add an equality check as well, just to make sure the correct time and tz are parsed.

technicat: Is there an example of a date check in the tests? I don't see any.

kkostov: We have some e.g. https://github.com/TootSDK/TootSDK/blob/main/Tests/TootSDKTests/ActivityTests.swift

kkostov:
@technicat Thank you for contributing this. Did you see something that would prevent it from being merged? I thought it was pretty cool to add.

technicat:
I just didn't want to spend any more time on the test.

kkostov:
I hear you, is it ok if I add the gets and merge it?

technicat:
sure!

kkostov:
Awesome, thanks!

Issue #103, opened 2023-03-06
Support for timeline streaming via server-sent events


bensyverson:
Hi all, based on the names of some of the API methods, I mistakenly thought TootSDK already supported [Mastodon's streaming API](https://docs.joinmastodon.org/methods/streaming/):

```swift
for await posts in try await client.data.stream(.timeLineHome) {
  ...
}
```

So I was super confused when I used TootSDK and it did not stream new posts. Digging into the code, it looks like "stream" refers to a Swift `AsyncStream`. And rather than calling the streaming endpoint for the user's home timeline (`GET /api/v1/streaming/use`), TootSDK calls the single-use endpoint (`GET /api/v1/timelines/home`).

For my purposes it's fine to periodically call `.refresh()`, but at some point it would be amazing to have a real async stream for a long-lived HTTP request or websocket.

Thanks for pulling together this SDK—it's nice!

kkostov:
Thanks for reporting this @bensyverson , that's a very cool insight.

Indeed, the streaming consists of facilitating the loading of different timelines. The `TootStream` is something we want to improve and this is an excellent venue to explore. There are challenges, of course, e.g. server-sent events are Mastodon-specific and not always available for other fediverse projects, so our solution should continue to allow for alternatives that can take over "under the hood".

Tunous:
Hey @kkostov, @davidgarywood I'm planning to use streaming endpoint to observe notifications https://docs.joinmastodon.org/methods/streaming/#notification and was wondering on how to add this to the sdk. 

Did you think about this topic previously and have any ideas on how it could be integrated. Should it be added to existing streams? Or something different?

davidgarywood:
I had planned to come back and integrate the mastodon streaming api under our data streams one, but hadn’t thought too deeply about it as it was in our kinda post 1.0 list. Well, we’re post 1.0 now 😃

I think, I’d love to see a draft PR with a new object, separate from the existing streams, and to then take a step back and see if they can be integrated easily or not.

It’ll either obviously slot in, or it’ll clearly need to be separate - but that shouldn’t be anyone’s concern when first pulling it together !

kkostov:
@Tunous are you using the current streams implementation in production?

Tunous:
@kkostov No, my app deals with timelines in non-standard way and the current streams implementation didn't work for my needs. I'm using TootClient.getTimeline(_:pageInfo:limit:) directly.

kkostov:
[...]

PR #281, opened 2024-05-30
Proof of Concept: Streaming via WebSocket

daprice:
Addresses #103 

I put together an implementation of WebSocket-based streaming using AsyncStream and Foundation’s `URLSessionWebSocketTask`. I tried to keep in mind the suggestions from the #103 discussion as much as possible, but ended up doing a few things differently. Feel free to use as much or little as you want from this PR 😄

## Caveats:
- I created a new `StreamingTimeline` type instead of using the existing `Timeline`. I didn’t think there was enough crossover between the set of timelines supported by the normal HTTP API and the streaming API to have any mapping between the two. E.g. for hashtag timelines, the HTTP API supports advanced queries with multiple hashtags, but the streaming API only supports subscribing to individual hashtags. Additionally, some of the timelines mean different things between the two APIs, like the home timeline in the streaming API is also used for notifications and server announcements.
- I didn’t use `TootStream` because I felt that the use cases for the normal HTTP API and streaming were different enough that they shouldn’t use the same abstraction. Additionally, there are a lot of potential complications that I think should be left up to each TootSDK user, such as when to transition between streaming and HTTP (e.g. a mobile app going into foreground/background states), how/whether to fill in gaps in the timeline if the streaming connection is unreliable, and so on.
- For now, this only supports Mastodon’s events and timelines. I think the others appear to be a superset of Mastodon's, so they should be able to be added fairly easily if TootSDK supports the payload types.

Inspired by how TootClient exposes both the more abstract `TootStream` and the more direct `getTimeline()`, I expose two different levels of abstraction for the streaming API: the lower-level `TootSocket` (a fairly thin wrapper around `URLSessionWebSocketTask` that provides an AsyncStream of all server-sent events), and the higher-level `StreamingClient` (itself built around `TootSocket`) that can share a single streaming connection between multiple subscribers and gracefully handles connection issues.

# TootSocket
This encapsulates `URLSessionWebSocketTask`, provides an AsyncThrowingStream of server-sent events in user-friendly types, and allows sending subscribe/unsubscribe requests. It’s up to the user of `TootSocket` how to handle multiple timeline subscriptions, handle any network/encoding/decoding errors, etc.

You can start an authenticated `TootSocket` by simply calling `tootClient.beginStreaming()`.

# StreamingClient
This is a higher-level API that handles multiple timeline subscriptions sharing a single connection. After a failed connection, it automatically retries (with an increasing delay after subsequent retries) up to a configurable maximum limit, in a way that is transparent to consumers.

To subscribe to a timeline, you just call `tootClient.streaming.subscribe(to: timeline)`, which returns an AsyncStream of both server-sent events for that timeline and connection status changes that the consumer can use to know when to revert to using the HTTP API. Disposing of the AsyncStream instance or cancelling the task that you are receiving events on will automatically unsubscribe from that timeline if there are no other consumers subscribed to the same timeline.

All calls to `tootClient.streaming.subscribe(to:)` share a single underlying `TootSocket` connection that the client attempts to maintain (automatically reconnecting if it gets disconnected) until you either call `tootClient.streaming.disconnect()` or the TootClient and its associated StreamingClient are deinited.

## Usage

Example code to keep a timeline view up to date via streaming, from a SwiftUI task modifier:
```swift
let stream = try await client.streaming.subscribe(to: streamingTimeline) // e.g. `.user`

// Check if the streaming connection is already up
// `isStreaming` is a state variable to show a UI element indicating streaming state and disable manual refreshing
isStreaming = await client.streaming.isConnectionUp

// Handle connection changes and received events
for await event in stream {
    switch event {
    case .connectionUp:
        isStreaming = true
    case .connectionDown:
        isStreaming = false
    case .receivedEvent(let eventContent):
        // switch over the possible enum cases and add posts to the view model, etc.
        handleStreamingEvent(eventContent, server: server)
    }
}
```

That’s all it takes! Multiple views can subscribe to the same timeline or different timelines and it’s handled seamlessly. Handling app backgrounding on iOS is also left up to the consumer, but all it takes (in SwiftUI, assuming one TootClient instance per Scene) is calling `tootClient.streaming.disconnect()` when the `scenePhase` changes to `.background`, and then restarting the above task when it returns to `.foreground`.


davidgarywood:
I love it @daprice - thank you!

One small bit of finessing might be to make EventContent Codable, and wrap up the logic in its internal int with the string comparison, inside of that. That's a real small nit though.

I really love the approach here and how this is laid out. I'm inclined for us to merge this straight in, so everyone can benefit  🙌

davidgarywood (approved):

kkostov (approved):
This is amazing, thank you so much 🤩!

daprice:
Fixed a few issues.

@davidgarywood The way Mastodon formats the event payloads is really weird, with the type of the `payload` value depending on the value of the `event` property, and some (but not all!) of the payload types have a second layer of JSON encoding to decode. I’m not sure how to handle that cleanly using Codable 😅

A couple other things:

- [ ] There are several places, especially in `StreamingClient`, where logging would be super helpful – should I go ahead with OSLog as mentioned in #172?
- [ ] Right now `StreamingClient` doesn’t actively check to see if the connection is alive. I need to do some more testing of how URLSessionWebSocketTask handles changes in network connectivity and might need to add a periodic ping to make sure it detects when the connection goes down.

davidgarywood:
> A couple other things:
> 
> * [ ]  There are several places, especially in `StreamingClient`, where logging would be super helpful – should I go ahead with OSLog as mentioned in [Log debug messages using OSLog logger #172](https://github.com/TootSDK/TootSDK/pull/172)?

I think great if you think it's doable and not too much time. If not then maybe come back for that after this PR?

daprice:
Logging should be good to go and I fixed a couple more bugs that I immediately caught with it. Would probably be good to let others try it and see if anyone has feedback. 🙂

One final caveat: Quinn from developer technical support has hinted that `URLSessionWebSocketTask` might eventually be deprecated ([forum post](https://developer.apple.com/forums/thread/731038?answerId=758875022#758875022)). I used it mainly for its simplicity and similarity with `URLSession` that’s already used throughout TootClient, and so far haven’t encountered any issues using it with Mastodon (knock on wood). That said, it shouldn’t be too difficult to replace it with another websocket library should the need arise, since only `TootSocket` and some of its supporting methods touch `URLSessionWebSocketTask`.

kkostov:
Looks ok to try it out I think, is this ready to merged or do you plan additional work @daprice ?

daprice:
@kkostov Should be ready, I think.

PR #285, opened 2024-07-02
Adds support for creating and updating filters

Tunous:
Ready for review.

Tested only via command line for now. I'll remove Draft once fully implemented and tested in my app in case I discover anything.


Tunous:
Should be good to go. I don't see anything missing now.

kkostov (approved):
Excellent! Thank you for making this work 🚀 !

PR #289, opened 2024-07-07
GoToSocial flavour

### LINKED ISSUE #287: GoToSocial support
I would like to add support for GoToSocial but unfortunately I encountered issue with how TootSDK detects flavors. Right now it's based on `/api/v1/instance` endpoint and version property. This works for other flavors but GoToSocial doesn't include its name here like other services.

For example https://gts.superseriousbusiness.org/api/v1/instance returns `0.16.0+git-f1cbf6f`

I searched for alternatives and found that there is a `/nodeinfo/2.0` endpoint with software.name property. This works for GoToSocial and Mastodon

https://gts.superseriousbusiness.org/nodeinfo/2.0
https://mastodon.world/nodeinfo/2.0

Now my question is how to resolve this properly in TootSDK. Should connection method be extended to call this endpoint after calling instance endpoint? This would always require 2 calls for Mastodon. Or should detection based on instance endpoint be fully replaced with nodeinfo check?

I would prefer the second option but am not sure whether this endpoint is a standard among all flavors.

Tunous:
`domainBlocks` feature was split to `adminDomainBlocks` since only this part is supported by GoToSocial.

Closes #287

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -92,6 +92,7 @@ let package = Package(
                 .copy("Resources/nodeinfo_catodon.json"),
                 .copy("Resources/nodeinfo_firefish.json"),
                 .copy("Resources/nodeinfo_friendica.json"),
+                .copy("Resources/nodeinfo_gotosocial.json"),
                 .copy("Resources/nodeinfo_iceshrimp.json"),
                 .copy("Resources/nodeinfo_mastodon.json"),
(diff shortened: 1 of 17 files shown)


davidgarywood (approved):
Nice one @Tunous  - thank you!

kkostov (approved):

PR #291, opened 2024-07-14
Prevent trailing new line from being added to extracted post plain text

Tunous:
While adding post editing I noticed that plain text always included trailing new line. This happened because `</p>` tag was always converted to a new line.

Fix makes it so the trailing `</p>` is skipped when converting to new lines.


technicat:
If I remember correctly there was also the opposite problem where it doesn't insert an extra line break between two paragraphs. In any case, fyi, I switched to this package (same author as the EmojiText package TootSDK uses)

https://github.com/divadretlaw/HTML2Markdown

Tunous:
@technicat Thanks I might consider it in future. For now I want to keep using existing parser.

kkostov (approved):

PR #295, opened 2024-07-27
Separate iceshrimp and catodon from firefish

Tunous:
This allows end clients to distinguish between these flavours. In my case I need this do display different icons and instance suggestions.

### DIFF
--- a/Sources/TootSDK/Models/Instance.swift
+++ b/Sources/TootSDK/Models/Instance.swift
@@ -224,11 +224,20 @@ extension Instance {
             return .akkoma
         }
         // 3.0.0 (compatible; Firefish 1.0.4-dev5)
-        // 4.2.1 (compatible; Iceshrimp 2023.12-pre3)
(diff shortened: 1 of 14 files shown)


technicat:
There's also

https://iceshrimp.dev/iceshrimp/iceshrimp.net

kkostov (approved):
Very nice! Thank you for submitting this PR.

kkostov:
@technicat is "iceshrimp.net" to be distinguished from "iceshrimp"?

technicat:
> @technicat is "iceshrimp.net" to be distinguished from "iceshrimp"?

You could go either way. It's a totally new codebase and the version string does say iceshrimp.net, so I like to distinguish it, but you could also just consider it a later version of iceshrimp, the misskey fork just has maintenance releases. The only differences I've marked down is lack of trending links/posts/tags but that could have changed by now, it's still under active development and there are not a lot of instances.

PR #296, opened 2024-07-30
Post editing fixes

Tunous:
The changes are tested by running live on account connected to pixey.org and venera.social instances.

For Friendica I've enabled post source endpoint as it is supported.

Pixelfed has more changes:
- Edit starts by concurrently updating media descriptions using separate endpoint since it is not possible to edit them using main endpoint.
- Content is sent as json instead of multipart
- Edited post is not returned from edit endpoint so it is requested manually
- Post source endpoint is marked as not supported

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Media.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Media.swift
@@ -67,13 +67,17 @@ extension TootClient {
             $0.url = getURL(["api", "v1", "media", id])
             $0.method = .put
 
-            let parts = mediaParts(
-                description: params.description,
-                focus: params.focus,
-                thumbnail: params.thumbnail,
-                mimeType: params.thumbnailMimeType
-            )
-            $0.body = try .multipart(parts, boundary: UUID().uuidString)
+            if flavour == .pixelfed {
+                $0.body = try .json(params)
+            } else {
+                let parts = mediaParts(
+                    description: params.description,
+                    focus: params.focus,
+                    thumbnail: params.thumbnail,
+                    mimeType: params.thumbnailMimeType
+                )
+                $0.body = try .multipart(parts, boundary: UUID().uuidString)
+            }
         }
         return try await fetch(MediaAttachment.self, req)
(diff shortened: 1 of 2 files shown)


Tunous:
One problem is that for Pixelfed there are now multiple potential points of failure. Catching error won't always mean here that the post has been edited. Either editing media descriptions could fail, main edit could fail, or even everything could succeed but then getting a resulting post could fail.

I didn't figure a way to resolve this while keeping single API for all flavours. But I think that's acceptable compromise, users can always retry.

kkostov:
Looks good to me @Tunous, I don't think there is a convenient way to improve this. Thank you for your PR.

technicat:
mitra and I think sharkey also require json for post edit (now I'm wondering about firefish, can't remember if I checked it), same with updating sheduled posts. Conveniently, mitra has an openapi (swagger) spec, as does gotosocial.

https://codeberg.org/silverpill/mitra/src/branch/main/docs/openapi.yaml

so I use multipart only for upload/updateMedia and updateCredentials

Tunous:
@technicat Did you test whether Mastodon works with json for edit post request too? If yes we can probably switch to it for all flavours.

technicat:
> @technicat Did you test whether Mastodon works with json for edit post request too? If yes we can probably switch to it for all flavours.

Yeah, I don't have editing of attachments  in my UI but otherwise it works. Btw there's also a similar return value issue with deletePost, a number of platforms don't have a return value (friendica, mitra, sharkey, firefish). So I have two versions of deletePost, but a simpler solution would be to just remove the return value assuming no one cares about it.

kkostov:
If removing the response makes it easier to support more flavours then that would be my preferred approach. I think if one really wants the edited post detail, they can get it back by id.

Tunous:
Updated per suggestion, now body is always sent as json. Also separated media description to separate endpoint for more flavours.

kkostov:
Awesome, will release it as soon as the CI clears :). Thank you so much @Tunous!

PR #297, opened 2024-08-01
Add latest preview card attributes returned by Mastodon

daprice:
Some recently added Mastodon features for link preview cards:
- article language
- publication date
- image description for article thumbnail
- authors array (in addition to the existing `author_name` and `author_url`) that can reference the author’s Fediverse account

These API additions are, as usual, missing from Mastodon’s documentation as far as I can tell, but were announced in [their blog post](https://blog.joinmastodon.org/2024/07/highlighting-journalism-on-mastodon/) and are live on mastodon.social. For example, [this post from the Verge](https://mastodon.social/@theverge_com/112887138012612832) has all of these properties, which you can test with this PR using `swiftyadmin get-post -u https://mastodon.social -i 112887138012612832 -t "[your_api_token]"`

All added properties are optional (they aren’t necessarily present on every post anyway), so there shouldn’t be any issues with servers that don’t send them.

### DIFF
--- a/Sources/TootSDK/Models/Card.swift
+++ b/Sources/TootSDK/Models/Card.swift
@@ -4,35 +4,43 @@
 import Foundation
 
 /// Represents a rich preview card that is generated using OpenGraph tags from a URL.
-public struct Card: Codable, Hashable {
+public struct Card: Codable, Hashable, Sendable {
     public init(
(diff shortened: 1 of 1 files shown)


kkostov (approved):
Awesome! Thank you!

PR #302, opened 2024-08-06
Add configurable behavior for retrying idle streaming connection

daprice:
Currently, if a streaming connection gets disconnected, StreamingClient will automatically retry even if the connection is idle (i.e. there are no subscribers to any streaming feeds). This PR makes the behavior in this case configurable, and by default does not automatically try to reconnect an unused connection.

The previous behavior of always retrying can be achieved by setting `retryIfNoSubscribers` to `true` on the StreamingClient instance.

## Use case

In my client app, I’m leaving the websocket connection open after the user finishes using it (e.g. by leaving whatever view is subscribed to a streaming feed) and just unsubscribing from the feed. This makes it faster and use less resources to resubscribe, **but** if the connection goes down while it’s not being used, having it always automatically reconnect right away is a waste of resources. This PR prevents that by making the default behavior to not reconnect unused connections after failure.

### DIFF
--- a/Sources/TootSDK/TootClient/Streaming/StreamingClient.swift
+++ b/Sources/TootSDK/TootClient/Streaming/StreamingClient.swift
@@ -67,6 +67,13 @@ public actor StreamingClient {
     /// This limit attempts to prevent excessive handshakes if the connection is available but unreliable.
     public var maxConnectionAttempts = 10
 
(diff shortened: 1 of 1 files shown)


kkostov (approved):

kkostov:
This is great, thanks 🤩!

PR #303, opened 2024-08-08
Add `TootFeature.searchExcludeUnreviewed`

daprice:
I broke out the mastodon flavor check if `excludeUnreviewed` is non-nil in `search()` into a TootFeature so API users can check to see if it’s supported using `TootClient.supportsFeature(.searchExcludeUnreviewed)` and adjust their behavior accordingly.


kkostov (approved):

kkostov:
Awesome! Thank you!

PR #307, opened 2024-08-12
Add streaming error

helje5:
Check the streaming event for error/status fields and decode those.


kkostov (approved):
Thank a lot @helje5, this adds nicely to #306!

PR #309, opened 2024-08-12
Fix rawValue of user notification timeline

### LINKED ISSUE #304: Decoding error, on errors w/ WebSocket against Mastodon.social
I'm playing w/ the WebSocket support. Along those lines:
```swift
      let socket = try await client.beginStreaming()
      try await socket.sendQuery(.init(.subscribe, timeline: .user))
      try await socket.sendQuery(.init(.subscribe, timeline: .userNotification))
      try await socket.sendQuery(.init(.subscribe, timeline: .direct))
```
I sometimes get various TLS failures, even though I pass in a URLSession that ignores such, but apart from that, if it goes through, it seems to trigger an error which results in:
```
(lldb) po error
▿ DecodingError
  ▿ keyNotFound : 2 elements
    - .0 : CodingKeys(stringValue: "stream", intValue: nil)
    ▿ .1 : Context
      - codingPath : 0 elements
      - debugDescription : "No value associated with key CodingKeys(stringValue: \"stream\", intValue: nil) (\"stream\")."
      - underlyingError : nil
```
This comes from:
```swift
extension StreamingEvent: Decodable {
    public init(from decoder: any Decoder) throws {
        let values = try decoder.container(keyedBy: CodingKeys.self)
        guard let timeline = StreamingTimeline(rawValue: try values.decode([String].self, forKey: .stream)) else {
```
Apparently in my case an error happens and has those keys:
- error
- status
So maybe that init should check for an optional .stream key, and if that's missing check for error&status and raise an appropriate exception.

daprice:
Fixes #304 - the rawValue of the .userNotification timeline was `userNotification` when it should have been `user:notification`

### DIFF
--- a/Sources/TootSDK/Models/Streaming/StreamingTimeline.swift
+++ b/Sources/TootSDK/Models/Streaming/StreamingTimeline.swift
@@ -75,7 +75,7 @@ extension StreamingTimeline: RawRepresentable {
         case .hashtag(let tag): ["hashtag", tag]
         case .localHashtag(let tag): ["hashtag:local", tag]
         case .user: ["user"]
-        case .userNotification: ["userNotification"]
+        case .userNotification: ["user:notification"]
         case .list(let listID): ["list", listID]
         case .direct: ["direct"]
         }


kkostov (approved):
Awesome! Thank you so much @daprice 🚀!

Issue #294, opened 2024-07-27
The Swift 6 Dungeon


kkostov:
> Dungeon is a general fantasy term for an underground adventuring area

Opening this issue to keep track of our progress and any discussions if needed during the transition. The initial plan is to:

1) Sort out warnings for Xcode 15.4 and Xcode 16 with no strict concurrency checking
2) Make adjustments to make Swift 6 with strict concurrency checks turned on

Apart from the main library, we seem to have a lot of concurrency warnings coming from the example iOS app. Unless a quick solution is possible there, removing the example and replacing it with docs is also a venue we can explore.

kkostov:
Together with @davidgarywood, we briefly discussed the pros and cons of keeping the [Vapor](https://github.com/TootSDK/TootSDK/tree/main/Examples/vaportoot) and [SwiftUI](https://github.com/TootSDK/TootSDK/tree/main/Examples/swiftui-toot) examples.

It seems the maintenance burden is greater than the added value of the examples themselves. By contrast, [the command line tool ](https://github.com/TootSDK/TootSDK/tree/main/Examples/swiftyadmin) offers a way to demonstrate TootSDK methods and functions in a concise way, focusing specifically on the library without distractions from surrounding boilerplate. So that will be the example to focus on going forward.

davidgarywood:
I've just stood up a draft PR and branch, that is an initial weedwhacking of things for strict concurrency here: https://github.com/TootSDK/TootSDK/pull/316

kkostov:
Hey folks, with Swift 6.2 and "approachable concurrency" we have a clean way to migrate the library. I think it would be easiest to take advantage of that (also means bumping Package.swift to 6.2 when it's released) rather than designing a complex solution for earlier toolchain versions. I know it's not ideal for backward compatibility but cutting support for older toolchains reflects how much time we can dedicate to this migration.

I'm eager to hear your thoughts! 

@technicat @daprice @Tunous @davidgarywood

daprice:
Same, I don’t see myself needing to use a newer version of the library with anything other than whatever the latest stable toolchain is at the time.

PR #316, opened 2024-08-22
Dev/dave swift6 concurrency wackamole

davidgarywood:
This is a draft branch, for fixes for Swift 6 and strict concurrency.

Please look at the commits, as they isolate some specific items and I suspect the last couple may well need modification to work well


davidgarywood:
Addresses #294

davidgarywood:
Very keen for @technicat  @Tunous @daprice  @helje5 and any others to weigh on this one also!

kkostov:
This looks really cool @davidgarywood, I will give it a good read :). Just note about MultipartKit, we needed to fork it because of the way they serialise collections is not how mastodon accepts them https://github.com/TootSDK/TootSDK/pull/62

Sources/TootSDK/TootClient/TootClient+Account.swift
```diff
@@ -282,6 +282,8 @@ extension TootFeature {
 
     /// Ability to edit your profile
     ///
+
+    @MainActor
```
Tunous, line 286: Any reason for this one? Might be problematic.

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
```diff
@@ -45,8 +43,6 @@
                             "There was a problem authenticating the user: no URL was returned from the first authentication step."))
                 }
 
-                authSession.presentationContextProvider = presentationContextProvider
```
Tunous, line 48: This is also interesting. From what I remember presentation context provider was always needed unless this changed or I misremember. What's the motivation behind this change?

Sources/TootSDK/TootClient/TootClient+Account.swift
davidgarywood: Giving it a global actor to hang off - this could be changed to something else. Keen for suggestions here!

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
davidgarywood: Lack of knowledge as to exactly what this was supporting within the original code, combine with the init of the anchor being @MainActor isolated. 

I’m definitely keen for testing and suggestions on this one !

Sources/TootSDK/TootClient/TootClient+Account.swift
Tunous: I would leave it without any actor. I don't see why accessing it should be limited to main actor, or any other actor.

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
helje5: I also recently did an own version of this and just marked the larger function MainActor to workaround the NS/UIWindow initialiser issue.
(BTW: I find it quite weird to return an empty window object as the anchor, even though it works :-) )

Sources/TootSDK/TootClient/TootClient+Account.swift
Tunous: I tried that and got no errors. Perhaps something changed since you added that annotation.

Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
Tunous: Compiled this myself and I had only error on defaultPresentationAnchor: Main actor-isolated default value in a nonisolated context.
Here is my attempt to keep the old functionality and resolve that problem. I initialize that default value if user did not provide their own provider and its not set yet. Since the whole function is isolated to MainActor it all works.

```patch
diff --git forkSrcPrefix/Sources/TootSDK/TootClient/TootClient+Account.swift forkDstPrefix/Sources/TootSDK/TootClient/TootClient+Account.swift
index 685004daee2902c3a1b59e9b388b9f36349b7138..83a643f9d9ba7c3eb4d6cf5bb8830890aba6a1c1 100644
--- forkSrcPrefix/Sources/TootSDK/TootClient/TootClient+Account.swift
+++ forkDstPrefix/Sources/TootSDK/TootClient/TootClient+Account.swift
@@ -40,7 +40,7 @@ extension TootClient {
     /// Update the user’s display and preferences.
     /// - Returns: The user’s own Account with source attribute
     public func updateCredentials(params: UpdateCredentialsParams) async throws -> Account {
-        try await requireFeature(.updateCredentials)
+        try requireFeature(.updateCredentials)
         let req = try HTTPRequestBuilder {
             $0.url = getURL(["api", "v1", "accounts", "update_credentials"])
             if self.flavour == .pixelfed {
@@ -282,8 +282,6 @@ extension TootFeature {
 
     /// Ability to edit your profile
     ///
-
-    @MainActor
     public static let updateCredentials = TootFeature(supportedFlavours: [
         .mastodon, .akkoma, .pleroma, .pixelfed, .firefish, .sharkey, .goToSocial, .catodon, .iceshrimp,
     ])
diff --git forkSrcPrefix/Sources/TootSDK/TootClient/TootClient.swift forkDstPrefix/Sources/TootSDK/TootClient/TootClient.swift
index f6a392687ca2e61152ade27da3412a6aa1d221a8..eb1d3afb2fd570b1851d6e3dd9354f009b0396d5 100644
--- forkSrcPrefix/Sources/TootSDK/TootClient/TootClient.swift
+++ forkDstPrefix/Sources/TootSDK/TootClient/TootClient.swift
@@ -52,6 +52,10 @@ public class TootClient: @unchecked Sendable {
     /// The current accessToken in use
     internal var accessToken: String?
 
+    #if canImport(AuthenticationServices) && !os(tvOS) && !os(watchOS)
+        internal var defaultPresentationAnchor: TootPresentationAnchor?
+    #endif
+
     /// Initialize a new instance of `TootClient` by optionally providing an access token for authentication.
     ///
     /// After initializing, you need to manually call ``TootClient/connect()`` in order to obtain the correct flavour of the server.
diff --git forkSrcPrefix/Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift forkDstPrefix/Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
index 33f3842fe4ab4343bdeed05366f140d1d7f8aa82..ebc5113f4e6d60e6f4edd82cb51fd1eb01e26097 100644
--- forkSrcPrefix/Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
+++ forkDstPrefix/Sources/TootSDK/TootClient/TootClient+ASWebAuthentication.swift
@@ -21,7 +21,14 @@
             prefersEphemeralWebBrowserSession: Bool = true,
             presentationContextProvider: ASWebAuthenticationPresentationContextProviding? = nil
         ) async throws -> String {
-            
+            var presentationContextProvider = presentationContextProvider
+            if presentationContextProvider == nil {
+                if defaultPresentationAnchor == nil {
+                    defaultPresentationAnchor = TootPresentationAnchor()
+                }
+                presentationContextProvider = defaultPresentationAnchor
+            }
+
             guard
                 let callbackURLScheme: String = URL(string: callbackURI)?.scheme
             else {
@@ -43,6 +50,8 @@
                             "There was a problem authenticating the user: no URL was returned from the first authentication step."))
                 }
 
+                authSession.presentationContextProvider = presentationContextProvider
+
                 authSession.prefersEphemeralWebBrowserSession = prefersEphemeralWebBrowserSession
                 authSession.start()
             }
@@ -53,4 +62,12 @@
         }
     }
 
+    // via: https://www.andyibanez.com/posts/using-aswebauthenticationaession-swiftui/
+    public class TootPresentationAnchor: NSObject, ASWebAuthenticationPresentationContextProviding {
+
+        public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+            return ASPresentationAnchor()
+        }
+    }
+
 #endif
```

davidgarywood: Brilliant, thanks @Tunous !

davidgarywood: This results in this warning on my side, with a clean build etc:

image

Adding the annotation as suggested results in this:

image

Tunous: Oh, I should note I used Xcode 16 Beta 6. There are other unrelated warnings but this class they marked as MainActor.

davidgarywood: Thanks @Tunous I'm running Xcode 15.4 here. Good to note.

I think I'd be happy to leave this with the warning, for Xcode 15.4 @kkostov - any thoughts?

kkostov: @davidgarywood how can I get this code on my side to test it out? 

From the screenshot, I think the suggestion of Xcode is not correct, the method should be marked as `nonisolated` instead. 

To answer your question, errors in Xcode 15.x are OK, as long as they're not errors in Swift 6 or errors with strict concurrency checking turned on in Xcode 15.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.7
+// swift-tools-version: 5.8
```
kkostov, line 1: I think we need to upgrade to 5.10 as the last pre-6 version to support from now on.

kkostov, line 1: I think we need to upgrade to 5.10 as the last pre-6 version to support from now on.

```suggestion
// swift-tools-version: 5.10
```

Package.swift
```diff
@@ -13,41 +13,32 @@ let package = Package(
     ],
     products: [
         // Products define the executables and libraries a package produces, and make them visible to other packages.
-        .library(
```
kkostov, line 16: I think we should revert this part.
I just tested with the latest version of multipart-kit - the serialisation is not correct for our purposes out of the box.
If MultipartKit requires concurrency-related modifications, we can pull their upstream and re-apply our tweaks (or re-visit doing this manually). Anyway, we can do this once the rest of TootSDK is cleared for concurrency.

PR #318, opened 2024-08-27
Add emoji reaction and unknown notification types

Tunous:
Adds emoji reaction notification type that might be sent by Sharkey.

Additionally, since this is a second time where someone reported non-working notifications due to a missing notification type, I decided to add dedicated case for unknown types. This way list of notifications will decode correctly giving clients chance to handle unknown cases in a better way. (Like displaying message to user to report that problem.)

### DIFF
--- a/Sources/TootSDK/Models/TootNotification.swift
+++ b/Sources/TootSDK/Models/TootNotification.swift
@@ -38,35 +38,119 @@ public struct TootNotification: Codable, Hashable, Identifiable, Sendable {
     /// Summary of the event that caused follow relationships to be severed. Attached when type of the notification is ``NotificationType/severedRelationships``.
     public var relationshipSeveranceEvent: RelationshipSeveranceEvent?
 
-    public enum NotificationType: String, Codable, Sendable, CaseIterable {
+    public enum NotificationType: Codable, Hashable, Sendable, CaseIterable, RawRepresentable {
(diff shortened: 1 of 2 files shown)


technicat:
Pleroma (and Akkoma and Mitra) also have an emoji reaction notification but it's pleroma:emoji_reaction

Tunous:
I pushed potential support for pleroma:emoji_reaction to https://github.com/TootSDK/TootSDK/commit/864c685c10f20ca12f5b163ca46416c4b42661e0

This required making NotificationType and structs that use it as only Decodable and not Encodable. Otherwise it would be encoded incorrectly for Pleroma/Akkoma. I'm not sure whether that is acceptable.

Alternatives that I can think of:
-pass in correct flavour to the encoder and access it in custom encode function for NotificationType. Probably userInfo on encoder? I didn't have a chance work with this api yet
- convert NotificationType to custom struct with rawValue and internal comparison logic. Public api would use emojiReaction and internally it would be split based on flavour. Will need a bit of experimentation though.
- add separate case for pleromaEmojiReaction

technicat:
I made all non-param definitions Decodable and params Encodable. It breaks some test code but I find it makes it clearer how everything is used.

kkostov (approved):
Awesome! Thank you!

PR #319, opened 2024-09-03
Fix user relationship endpoint for sharkey

Tunous:
Sharkey returns smaller set of data than Mastodon when fetching user relationship. To properly decode its responses additional fields needed to be marked as optional.

I considered adding default values instead of optional but ultimately decided that it'll be better to leave that up to clients.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -79,6 +79,7 @@ let package = Package(
                 .copy("Resources/mastodon.json"),
                 .copy("Resources/pleroma.json"),
                 .copy("Resources/relationship.json"),
+                .copy("Resources/relationship_sharkey.json"),
                 .copy("Resources/scheduled_post_sensitive.json"),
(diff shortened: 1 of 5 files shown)


kkostov (approved):
Awesome, thank you!

Tunous:
Testing it a bit further I now see that it always returns the same set of details for me… No matter the account I test against. Before I assumed that it's correct because when I followed someone it returned correct info. But now I see it was only as a response to follow call. Something else must be incorrect

@technicat Is this endpoint working for you correctly?

technicat:
yeah I'm seeing the same thing

technicat:
I can't believe I haven't noticed this problem before (maybe it was introduced recently? I think there's been some rewriting of their Mastodon API in progress).

https://activitypub.software/TransFem-org/Sharkey/-/issues/557

PR #320, opened 2024-09-23
Add optional parameter for client website

mikakruschel:
Clients can provide a URL to their website, which is the link when pressing on the client name in the details of a toot. (https://docs.joinmastodon.org/methods/apps/#form-data-parameters)
This PR exposes this parameter, which has been "" by default and still is the default, to the client.

### DIFF
--- a/Sources/TootSDK/Models/CreateAppRequest.swift
+++ b/Sources/TootSDK/Models/CreateAppRequest.swift
@@ -7,5 +7,5 @@ struct CreateAppRequest: Hashable, Codable {
     let clientName: String
     let redirectUris: String
     let scopes: String
-    let website: String
+    let website: String?
 }
--- a/Sources/TootSDK/TootClient/TootClient+Request.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Request.swift
@@ -64,14 +64,14 @@ extension TootClient {
     internal func getAuthorizationInfo(
         callbackURI: String,
         scopes: [String],
-        website: String = "",
         responseType: String = "code"
     ) async throws -> CallbackInfo {
 
         let createAppData = CreateAppRequest(
             clientName: clientName,
             redirectUris: callbackURI,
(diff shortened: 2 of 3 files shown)


technicat:
You might also want to make it an optional, if you're changing it, e.g. at one time Iceshrimp.net was checking any non-null website and throwing an exception if it wasn't a valid URL (though they've changed that accomodate tootsdk clients)

https://chat.iceshrimp.dev/#narrow/stream/1-general/topic/.2ENET.20clients/near/12061

mikakruschel:
I'm not sure if this is really necessary, as so far, an empty string has always been sent. (The link you provided is not publicly accessible.)

Tunous (approved):

technicat:
sorry, here's the iceshrimp chat post. they've changed the code since then to treat empty string same as null, but note also that although the mastodon spec doesn't list website as optional, the return value sample does show a null website.


> I found that any iOS app that uses the [TootSDK](https://github.com/TootSDK/TootSDK) doesn't work (Piplio, [FediCat](https://github.com/technicat/fedicat)). Error is 400, Authorization Failed immediately after entering the instance url.

>Thanks for testing and reporting this! We talked about this yesterday but for the readership here the tootsdk problem turned out to be that it's supplying a non-null website value (an empty string), which triggers this code in [iceshrimp.net](http://iceshrimp.net/):

```csharp
if (request.Website != null)
            try
            {
                var uri = new Uri(request.Website);
                if (!uri.IsAbsoluteUri || uri.Scheme is not "http" and not "https") throw new Exception();
            }
            catch
            {
                throw GracefulException.BadRequest("Invalid website URL");
            }
```

mikakruschel:
Oh, you're right—it should be optional. I can easily change that. I'm wondering whether we should keep the current default value of `""` to avoid introducing a breaking change, or change it to `nil`, since technically `""` isn't valid.

davidgarywood (approved):
Nice one !

Issue #317, opened 2024-08-22
Instance v2


daprice:
Mastodon’s [Instance v2](https://docs.joinmastodon.org/methods/instance/#v2) API provides a fair amount of info that isn’t in v1, like MAU, a link to the server status page, accurately named properties (`uri` in v1 wasn’t actually a URI, renamed to `domain` in v2), more configuration info, a higher resolution version of the server banner with blurhash, and most recently the URLs of various sizes of the favicon. They’ve marked v1 as deprecated and it looks like anything new they add will be v2-only.

I’m assuming TootSDK went with v1 for compatibility with other instance flavors, so I have no idea what the right approach is here. Just at a glance, support for v2 in various flavors appears to be all over the place:

| software | supports v2? | notes |
| --- | --- | --- |
| Mastodon | yes | they just added an `api_version` property that might be helpful here, as a v2 exclusive: https://github.com/mastodon/mastodon/pull/31354 |
|Pixelfed | yes |  |
| Pleroma | yes | has its own additions in a `pleroma` property, but missing the thumbnail blurhash and 2x resolution |
| Friendica| yes | has its own additions in a `friendica` property, also missing the extra thumbnail versions |
| GoToSocial | yes | has its own additions mixed in (& I think it might be using the `domain` property differently than Mastodon?), also missing extra thumbnail versions |
| Firefish | yes | missing the extra thumbnail versions |
| Akkoma | no | v2 returns 404  |
| Sharkey | no | v2 returns 404 |
| Iceshrimp | no | v2 returns 404 |

There are also some properties that are in v1 but not v2, like detailed stats, further complicating things. That said, I’d find the v2 exclusives like the high res banner and favicons very useful.

technicat:
If you want to add Instance V2 you can start with what I have. I've made a ton of changes overall so there will be some incompatibilities (I use an enum for language strings, only Decodable for non-param models...) but you can tweak or stub those out (and it's also still missing some fields like blurhash). As for where to use it, like you say it's rather haphazard - my platform detection usually starts out with checking the nodeinfo for the platform and version (even on Mastodon as I recall it's only on 4.0 or later) and then I bring up instance v1 or v2 as applicable.

https://codeberg.org/technicat/tootsdk/src/branch/main/Sources/TootSDK/Models/Instance/InstanceV2.swift

technicat:
[...]

PR #321, opened 2024-10-22
Instance v2

daprice:
Implements #317 

`tootClient.getInstanceInfo()` now tries to use the highest supported API and returns a struct conforming to the `Instance` protocol. This protocol contains fields that are present and equivalent across both versions. Where v1 and v2 use the same property name in conflicting ways, I left those properties out of the protocol (e.g. I didn’t include `description` because it contains the long description in v1 and short description in v2; likewise `configuration` contains different items in each version). This does mean that for version-dependent properties, users have to downcast to `InstanceV1` or `InstanceV2`. Otherwise it’s fairly similar to @technicat’s implementation linked in the issue thread.

For use cases where you want to manually call a particular version of the API and get a concrete type, I’ve made `getInstanceInfoV1()` and `getInstanceInfoV2()` methods public on TootClient.

### Migrating client code
This is a breaking change, but if you currently only access properties that are present in the `Instance` protocol, you may not need to make any changes.

#### Continuing with v1 only
If you only need v1, the existing `tootClient.getInstanceInfo()` was renamed to `tootClient.getInstanceInfoV1()` and the existing `Instance` struct was renamed to `InstanceV1`.

#### Going v2 “only”
If you only want to deal with the V2 API, but still want to support instance flavors that only support v1, you can use `getInstanceInfo()` to use the latest supported version, then chain with `.v2Representation()` to convert all resulting structs to `InstanceV2`. If the server supports V2, this simply returns the struct as-is, cast to `InstanceV2`. If the server returns a V1 instance, this creates an `InstanceV2` with unsupported fields set to nil, and converts nested structs like `configuration` to the v2 format.

I used this approach for swiftyadmin’s registration check: https://github.com/daprice/TootSDK/blob/instancev2/Examples/swiftyadmin/Sources/swiftyadmin/Account/RegisterAccount.swift#L31

## Questions around flavor-specific extensions
I’m wondering what would be the best way to handle flavor-specific extensions to the Instance struct. Pleroma and Friendica put all of their additions under a `pleroma`/`friendica` property, which could be neatly handled with an optional nested struct for each one (this is how @technicat’s code linked in the issue thread works). However, GoToSocial added extra properties directly to their Instance object and nested within the `configuration` struct, which doesn’t fit neatly into this approach.

My ideas for handling GoToSocial would be either:
1. Just add the extra properties as optionals on `InstanceV2`, `InstanceConfiguration` and so on, and note in the documentation which flavor each property corresponds with. I’d worry that this could eventually result in a collision if different flavors use the same property name for different things.
2. Move GTS’s extra properties to a nested `GoToSocial` struct during decoding in its Codable implementation.
3. Make `InstanceV2` itself a protocol, with `PleromaInstanceV2`, `FriendicaInstanceV2`, `GoToSocialInstanceV2` etc as conforming types. The same would have to be done for the nested types in `InstanceConfiguration`. This would make it clear which properties belong to which flavor, prevent conflicts, and preserve each flavor’s "native" structure, but would require changes to flavor detection and an extra downcast for SDK users.

## Possible next steps
Mastodon added an `api_versions.mastodon` property in InstanceV2 with its own version number, which they [explain as](https://docs.joinmastodon.org/entities/Instance/#api_versionsmastodon):

> API version number that this server implements. Starting from Mastodon v4.3.0, API changes will come with a version number, which clients can check against this value.

Other instance flavors could implement this to indicate which parts of the Mastodon API they support, as well as advertise their own API additions to clients. Would this be useful to incorporate directly into TootClient, for use in `supportsFeature()` for example?


kkostov:
Thanks a lot for this PR @daprice, this is great work! I'm happy to merge it.
To answer some of your questions: 

## around flavor-specific extensions

After a chat with @davidgarywood, I think it will be best to indeed adopt the protocol-per-flavour-version approach (3). This will be a one-time refactor to move flavour specific logic to dedicated classes. I will dedicate some time in the next day to PR it in one go.

## api_versions.mastodon

With (3) in place, we can adopt Mastodon specific functionality in the future. For the moment, I don't see added value in having this before we know more on how Mastodon plans to use this version. Our flavours already include "the general instance version" as a metric when considering feature support. If this field becomes important for mastodon, it can be included in the flavour/feature detection.

kkostov (approved):

PR #323, opened 2024-11-07
Attributed string renderer

Tunous:
This is a replacement renderer for `UIKitAttribStringRenderer` and `AppKitAttribStringRenderer` that uses newer `AttributedString` API. 

Changes from old renderers:
- Fixed bold and italic styles
- Added styling of headings
- Improved how new lines are added between paragraphs and inline elements
- Plain string now has consistent new lines with attributed string
- Nested lists have increased indent
- Emojis are ignored for now

This is an initial version that should be good for most of the cases. I'm looking for feedback if anyone decides to use it and encounters posts that would render incorrectly.

Here is an example of a [markdown formatted post](https://poa.st/@pipilo/posts/Ano93ccHOUVMJPq0h6) from pleroma as it appears in my app using this renderer:
Screenshot 2024-11-07 at 19 12 34

### DIFF
--- a/Sources/TootSDK/HTMLRendering/AttributedString+Util.swift
+++ b/Sources/TootSDK/HTMLRendering/AttributedString+Util.swift
@@ -1,0 +1,73 @@
+//
+//  AttributedString+Util.swift
+//  TootSDK
+//
+//  Created by Łukasz Rutkowski on 06/11/2024.
+//
+
+import Foundation
+
+#if canImport(UIKit)
+    import UIKit
+
+    typealias _Font = UIFont
+#elseif canImport(AppKit)
+    import AppKit
+
+    typealias _Font = NSFont
+#endif
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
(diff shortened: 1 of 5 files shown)


Tunous:
Looks like CI is failing

- Lint: No idea what is wrong here
- Build: I used Testing library available with Xcode 16, should I convert to XCTest or can CI be updated somehow?

kkostov:
Regarding the build, it's failing because we're building with Xcode 15 and so Swift Testing is not available there. I haven't had a chance to get into Swift 6+ stuff but if the time has come, we can drop Xcode 15. What do you think @davidgarywood?

davidgarywood:
> Regarding the build, it's failing because we're building with Xcode 15 and so Swift Testing is not available there. I haven't had a chance to get into Swift 6+ stuff but if the time has come, we can drop Xcode 15. What do you think @davidgarywood?

Happy for us to bump things to Xcode 16 on the CI if we can!

kkostov:
Okay, will drop Xcode 15 on main so this can build.

PR #324, opened 2024-11-12
Fix configuration.urls not working with the Instance v2 API

daprice:
Somehow I missed that the instance v2 API uses a different property name for the streaming URL. The result was that you couldn’t start streaming if TootSDK used v2 to get the instance info. This fixes that by using a separate type for the URLs struct for v1 and v2. That means this fix is a breaking change.

For consistency, I also replaced the `Instance` protocol’s `urls` property with a `streamingURL` property that returns the appropriate value from either version, and in v2 I added Mastodon's undocumented `status` property that returns the URL of the server status page.


kkostov (approved):
Awesome, thank you!

PR #325, opened 2024-11-15
Pleroma fixes

Tunous:
[Fix error when trying to follow account without extra params](https://github.com/TootSDK/TootSDK/commit/f8d5b75f47ab55e70e0b39150769c7bfed0f5795)
[Fix decoding of card with width/height as string instead of int](https://github.com/TootSDK/TootSDK/commit/67880f5eeb8db42f9ff34dc563316a87dfbde618)


kkostov (approved):
Awesome! Thank you 💜 !

PR #328, opened 2024-11-16
Fix broken pagination if links share the same parameter names

### LINKED ISSUE #326: Broken pleroma pagination
Given a request for home timeline with `min_id` Pleroma responds with different pagination links than Mastodon. Mastodon responds with `max_id` in link for next page and `min_id` in link for previous page. Pleroma instead includes both `max_id` and `min_id` for both links. This breaks because TootSDK extracts first encountered query value, no matter in which link it was received.

I'm raising this as a issue instead of pull request because I'm not entirely sure how proper fix should work. Should min_id be extracted only from prev link, or maybe both should be returned for user to decide? Or maybe something entirely different?

Ultimately what I want to have is:
When I request a timeline, in response I want to receive posts and information what to send to receive previous or next page.

Tunous:
Closes #326 

Fixes incorrect pagination identifiers if there are duplicate parameters in next and prev links.

For compatibility `info` in `PagedResult` is filled with assumption that ids will be in specific places, and `nextPage` with `previousPage` are initialized as inverted variants of next/prev links.

### DIFF
--- a/Sources/TootSDK/HTTP/PagedResult.swift
+++ b/Sources/TootSDK/HTTP/PagedResult.swift
@@ -5,36 +5,22 @@ import Foundation
 
 public struct PagedResult<T: Decodable>: Decodable {
     public let result: T
-    public let info: PagedInfo
+
+    /// Returns pagination object for the next page of results if one is available. A next page contains results newer than the current page.
+    public let nextPage: PagedInfo?
+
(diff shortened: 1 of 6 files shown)


Tunous:
I used this for some time in production already and it works correctly. Should be good to merge.

davidgarywood (approved):
This is looking neat - thank you!

kkostov (approved):

PR #329, opened 2024-11-24
Get account preferences and gif autoplay setting

daprice:
1. Added Mastodon’s “autoplay gifs” setting to `Preferences` (see mastodon/mastodon#22706 - there’s currently only a way to read this setting but not write it, so its usefulness for clients is questionable, but it still seems important to include since it’s accessibility-related)
2. Added a TootClient method to get the Preferences (I didn’t see an existing method for this)
3. Added `get-preferences` subcommand to swiftyadmin for testing this functionality.


kkostov (approved):
Awesome, thank you!

PR #330, opened 2024-12-07
Add notification annual_report type

Tunous:
New notification type added in https://github.com/mastodon/mastodon/pull/32709


davidgarywood (approved):
Nice one !

PR #332, opened 2025-02-04
Add methods to get privacy policy and terms of service for an instance

daprice:
The `PrivacyPolicy` struct here is identical to `ExtendedDescription`, but Mastodon defines it as a separate entity type, so this PR follows that.

PrivacyPolicy was added to Mastodon in 4.0.0 but not documented at the time; TermsOfService is upcoming in 4.4.0 but theoretically already supported on mastodon.social, although they haven’t set their terms yet so it currently returns `404`.


technicat:
I just noticed the return struct is identical to ExtendedDescription, so could reuse that.

daprice:
> I just noticed the return struct is identical to ExtendedDescription, so could reuse that.

They’re defined separately from ExtendedDescription in the Mastodon codebase so it’s possible they might diverge in the future. This avoids the need for breaking changes if/when they do.

PR #338, opened 2025-02-14
Add `reason` parameter to `TootSDKError.invalidParameter`

### LINKED ISSUE #334: Add reason for ScheduledAt Parameter error
When scheduling posts, we proactively check on client side that the schedule time must be at least 5 minutes in the future which some flavour servers enforce. However, the message is quite confusing at the call site. We should find a way to provide more context back in the exception.

cf https://github.com/TootSDK/TootSDK/blob/c7795f4971a1d703323acd2c388a171a5fbfc35c/Sources/TootSDK/Models/ScheduledPostRequest.swift#L58

finestructure:
Fixes #334 

As [discussed on the 'don](https://toot.iamkonstantin.eu/@konstantin/113996972469397165), hope that helps!


kkostov:
Brilliant! Thanks a lot @finestructure 🚀!

kkostov (approved):

davidgarywood (approved):

PR #339, opened 2025-02-16
Add missing MediaAttachment properties

daprice:
- `previewRemoteUrl` property on `MediaAttachment`
  - from mastodon/mastodon#14358
- `colors` metadata used in Mastodon to theme the player for audio attachments using colors extracted from the artwork
  - from mastodon/mastodon#14209

### DIFF
--- a/Sources/TootSDK/Models/MediaAttachment.swift
+++ b/Sources/TootSDK/Models/MediaAttachment.swift
@@ -60,7 +60,7 @@ public struct AttachmentMeta: Codable, Hashable, Sendable {
     public var original: AttachmentMetaInfo?
     public var small: AttachmentMetaInfo?
     public var focus: AttachmentMetaFocus?
-    /// For ``AttachmentType/audio`` attachments, contains colors extracted from the thumbnail.
+    /// For ``AttachmentType/audio`` attachments, contains colors matching the artwork (thumbnail) associated with the audio file.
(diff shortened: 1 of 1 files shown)


Sources/TootSDK/Models/MediaAttachment.swift
```diff
@@ -56,6 +60,8 @@ public struct AttachmentMeta: Codable, Hashable, Sendable {
     public var original: AttachmentMetaInfo?
     public var small: AttachmentMetaInfo?
     public var focus: AttachmentMetaFocus?
+    /// For ``AttachmentType/audio`` attachments, contains colors extracted from the thumbnail.
```
kkostov, line 63: Is `/audio` correct here?

daprice: Yes, at least with mastodon the `colors` are only present for audio and match the artwork associated with the audio file (like album art). I just used the terminology from the mastodon PR that added the feature, is there a different way of saying it that would be clearer?

kkostov: Oh I see! Something like this?

```suggestion
    /// For ``AttachmentType/audio`` attachments, contains colors matching the artwork (thumbnail) associated with the audio file.
```

kkostov (approved):
Awesome, thanks a lot @daprice ✨

PR #340, opened 2025-02-23
Decode dates formatted as unix timestamp

Tunous:
This is a fix for instances on https://gblog4.popolon.org/snac/, which appears to be formatting marker update dates as unix timestamp.


kkostov (approved):
Thank you @Tunous 🤩!

PR #342, opened 2025-04-29
Fix `ExpandMedia` expecting camel case when the Mastodon API returns snake case

daprice:
Previously, calling `TootClient.getPreferences()` would throw an error if the user had their Mastodon media setting set to “show all” or “hide all”. The API returns the strings `hide_all` or `show_all`, but TootSDK expected `hideAll` or `showAll`


kkostov (approved):
Nice! Thank you @daprice

Issue #343, opened 2025-04-29
Handle unknown enum cases returned from API


daprice:
Currently TootSDK's enums throw an error when decoding if the server returns an unknown value. I don’t think this is ideal because API updates frequently add new values.

A good example is `Filter.Action`, where Mastodon recently added a new `blur` case. Their [docs](https://docs.joinmastodon.org/entities/Filter/#filter_action) make it clear that clients should expect new values to be added and fall back to the same behavior as `warn` when receiving an unknown value. It’s an easy fix for TootSDK clients to just add the extra case to that enum, but it still means that until every client is updated, every API call that returns a `Filter` will error out.

I’m thinking every decodable enum in TootSDK should probably be able to handle unknown values gracefully. There are a few approaches detailed [in this blog post](https://5sw.de/2022/01/unknown-enum-cases/) – I like their `UnknownDecodable` protocol idea. I’d also suggest taking a cue from how Swift Optionals are implemented by defining a wrapper enum like:

```swift
@frozen public enum OpenEnum<Wrapped: Codable>: Codable {
    case some(Wrapped)
    case unknown(String)

    init(from: Decoder) {
        // Try to decode `Wrapped` and if it fails, return `.unknown` with the raw string
    }
}
```

and then use that for every occurrence of a decodable enum in TootSDK:

```swift
// Sources/TootSDK/Models/Filter.swift

public struct Filter … {
    …
    public var filterAction: OpenEnum<Action>
    …
}
```

This kind of wrapper has the bonus of distinguishing between “the server literally returned `unknown`” (as seen in `MediaAttachment.type` where “unknown” is a valid string for the server to return) and “the server returned a value that is unknown to TootSDK”, which I think is an important distinction.

Anyone else have any other suggestions or thoughts?

Tunous:
I would prefer flat hierarchy over additional wrapper. The unknown case should appear only if TootSDK is not updated, which ideally would be never. Having a single switch is more convenient.

In situations where Mastodon returns case named "unknown" the fallback case can be named differently.

kkostov:
I like the approach of having a default value when the response is not supported by TootSDK (potentially with access to the raw value as string) as it will make it explicit and allow consuming apps to handle the case instead of having the entire request fail. 

Perhaps we can tweak the name so it's clear it's "unknown to TootSDK"? e.g. an enum case with associated value:

`unparsedByTootSDK(originalValue: String)`

Tunous:
[...]

PR #345, opened 2025-05-01
Add “blur” filter action

daprice:
This was added in Mastodon 4.4.0 (Mastodon API version 5); see Mastodon/Mastodon#34256

Pending a solution for #343, clients must update ASAP with this fix to avoid errors when making a wide range of API calls if the user has set up any filters with the new option on Mastodon servers running nightlies (e.g. mastodon.social).


kkostov (approved):
🎉

PR #348, opened 2025-05-13
Add instance domain blocks and peers endpoints

daprice:
Adds support for `api/v1/instance/peers` (list of connected servers) and `api/v1/instance/domain_blocks` (publicly viewable list of moderated servers).

I called the new struct `InstanceDomainBlock` because TootSDK already used `DomainBlock` for what Mastodon calls `Admin:DomainBlock`. Let me know if you’d rather shuffle the names around to match those used by the Mastodon documentation (i.e. rename TootSDK’s existing `DomainBlock` to `AdminDomainBlock` and the new `InstanceDomainBlock` to just plain `DomainBlock`).


kkostov (approved):
This is awesome! Good call on naming it `InstanceDomainBlock`, it's better to avoid a breaking change, and it feels like they'd need clarification either way.

PR #350, opened 2025-05-30
Support versioned/dated terms of service endpoint

daprice:
Announced in https://blog.joinmastodon.org/2025/05/legal-features-updates/

`getTermsOfService()` now returns a `TermsOfService` struct instead of `PrivacyPolicy`. This is a breaking change from my previous PR #332 (but only for the terms of service endpoint, not the others) because Mastodon changed what the endpoint returns since then. There shouldn't be any servers in the wild that still return the `PrivacyPolicy` type because that was only ever in nightly builds and not an official release.


kkostov (approved):
Nice! Thank you :)

PR #352, opened 2025-06-19
Safe decoding of enums

### LINKED ISSUE #343: Handle unknown enum cases returned from API

Tunous:
Implemented as `OpenEnum`. The `UnknownDecodable` protocol didn't work due to incompatibility of enums with raw value and enums with associated parameters.

For convenience, added the `value` property that allows accessing the underlying enum as an optional value. It returns nil if enum was decoded to an unknown value.

Only enums that are decoded are wrapped in `OpenEnum`. Initializers and enums used only in sent parameters are not affected. In situations where there would be a new known case for them, they should be contributed to TootSDK instead of using the unparsed case.

Closes #343

### DIFF
--- a/Examples/swiftyadmin/Package.resolved
+++ b/Examples/swiftyadmin/Package.resolved
@@ -68,8 +68,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/scinfu/SwiftSoup.git",
       "state" : {
-        "revision" : "f707b8680cddb96dc1855632340a572ef37bbb98",
-        "version" : "2.5.3"
+        "revision" : "aa85ee96017a730031bafe411cde24a08a17a9c9",
+        "version" : "2.8.8"
       }
     }
   ],
--- a/Sources/TootSDK/HTTP/OpenEnum.swift
+++ b/Sources/TootSDK/HTTP/OpenEnum.swift
@@ -1,0 +1,70 @@
+//
+//  OpenEnum.swift
+//  TootSDK
+//
+//  Created by Łukasz Rutkowski on 17/06/2025.
+//
+
+import Foundation
+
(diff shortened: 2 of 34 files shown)


kkostov (approved):
Thank you for this contribution @Tunous, this is fantastic work!

PR #355, opened 2025-07-18
Quote posts

daprice:
This supports the quote post API features from [Mastodon 4.4](https://blog.joinmastodon.org/2025/07/mastodon-4-4-for-devs/).

The Mastodon API is kind of weird here: the `quote` property can contain either a Quote or ShallowQuote object that are almost the same except for a property that can be populated with either the ID of a post, the full post content, or nil (if nil, there is no way to distinguish between the shallow and regular types using Codable, that I can see?). I just made one Quote struct with an optional enum that either contains a Post or ID (String) depending on which property is present, so there’s only one optional to unwrap within the Quote and no type casting needed. I also added tests for decoding the examples of each type provided in the Mastodon docs.

I plan to submit more PRs for the rest of the Mastodon 4.4 changes in that blog post over the next few days, in case you want to wait before creating a release.


daprice:
Good catch @technicat. @kkostov @davidgarywood do you have a preferred approach for handling this?

Seems like the cleanest way (for the API user and for decoding) would be to make `Quote.state` optional, and if the `post.quote` field contains a `post`, decode it to a `Quote` with a nil `state` and the Post in the `Quote.quotedPost` property.

Tunous:
@daprice Instead of optional it could be accepted state. Should fit since this is the only state where we are given a non-nil post.

daprice:
I wanted to avoid setting “accepted” when we don’t receive that from the server, since it indicates that the quoted person gave consent when, AFAICT, Akkoma etc don’t have a mechanism for that.

kkostov (approved):
Thank you @daprice, this looks perfect!

PR #356, opened 2025-07-18
Support Mastodon 4.4 account management additions

daprice:
#### API support:
- [X] added support for endpoints to endorse (feature), unendorse, and query endorsements (featured accounts) by account ID
- [X] support getting attribution domains using verifyCredentials and setting them using updateCredentials

#### swiftyadmin:
- [X] added EndorseAccount, UnendorseAccount, GetEndorsements, GetOwnAccountEndorsements commands
- [X] added VerifyCredentials command that prints CredentialAccount `source`

#### Notes
- Calling updateCredentials with the `attributionDomains` property is only supported if the server supports Mastodon API version 3 or higher.
- The `endorse` and `unendorse` endpoints are intended to replace the older `pin` and `unpin` which are now deprecated in Mastodon 4.4. The docs don’t give an API version for this change, just the main Mastodon version number.
- Since TootClient doesn’t (currently) track the Mastodon API version or Mastodon version, I didn’t add a version check in any of these methods, I just noted the requirement and deprecation in the DocC comments of each method.
- The [Mastodon 4.4 for developers](https://blog.joinmastodon.org/2025/07/mastodon-4-4-for-devs/) post also notes new endpoints for featuring/unfeaturing hashtags, which seem to be near-duplicates of existing endpoints that TootSDK already supports. (update: in addition to returning a `Tag` instead of a `FeaturedTag`, the new endpoints are intended to be easier to use by not requiring the tag name in form data.) Since TootSDK already has an implementation for the existing endpoints with wider compatibility, I left these alone.

### DIFF
--- a/Examples/swiftyadmin/Sources/swiftyadmin/Account/EndorseAccount.swift
+++ b/Examples/swiftyadmin/Sources/swiftyadmin/Account/EndorseAccount.swift
@@ -1,0 +1,30 @@
+//
+//  EndorseAccount.swift
+//  swiftyadmin
+//
+//  Created by Dale Price on 7/17/25.
+//
+
+import ArgumentParser
+import Foundation
+import TootSDK
+
+struct EndorseAccount: AsyncParsableCommand {
+
+    @Option(name: .short, help: "URL to the instance to connect to")
+    var url: String
+
+    @Option(name: .short, help: "Access token for an account with sufficient permissions.")
+    var token: String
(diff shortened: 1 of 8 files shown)


kkostov (approved):
Thanks @daprice this looks good!
I opened a separate issue to introduce the API version checks with a possible solution, I'd appreciate your thoughts on the proposed design.

PR #360, opened 2025-07-26
Expose HTTP response headers from API calls

### LINKED ISSUE #331: Expose HTTP response headers from API calls
Currently, none of the TootClient.get*X* methods expose the HTTPResponse or its headers to the caller or in any other way, except for the ones that use `fetchPagedResult` which reads the `Link` header but discards the rest.

Mastodon returns useful information via response headers, including the referrer policy (which is needed to determine whether clients should send a `referer` header when loading attachments) and info about rate limits. It would be useful if the client could access these in some way.

kkostov:
This PR fixes #331 by exposing HTTP response headers and metadata from TootSDK API calls while maintaining backward compatibility. 

### 1. TootResponse<T> Wrapper Type

Create a new generic wrapper type that contains:

- **data**: The decoded response data (existing return type)
- **headers**: HTTP response headers as `[String: String]`
- **statusCode**: HTTP status code
- **url**: Response URL
- **rawBody**: Raw response body as `Data` (for debugging/fallback)

### 2. Backward Compatible Implementation

- **Existing method signatures remain unchanged** - No breaking changes to current API
- **Implementation migration** - Current method implementations will be moved to new `*Raw()` methods
- **Delegation pattern** - Existing methods will be refactored to call Raw methods and extract only the data property

### 3. Method Naming Convention

For each existing method like `getLists()`, add:

- `getListsRaw() async throws -> TootResponse<[List]>`

## Before merging

- [ ] Update CLAUDE.md to reflect the *Raw() method approach for future additions.

### DIFF
--- a/Sources/TootSDK/TootClient/TootClient+Media.swift
+++ b/Sources/TootSDK/TootClient/TootClient+Media.swift
@@ -58,29 +58,54 @@ extension TootClient {
     /// - Parameter id: The local ID of the attachment.
     /// - Returns: `Attachment` with a `url` to the media if available. `nil` otherwise.
     public func getMedia(id: String) async throws -> MediaAttachment? {
-        do {
-            let response = try await getMediaRaw(id: id)
-            return response.data
-        } catch TootSDKError.invalidStatusCode(let data, let httpResponse) {
-            // Mastodon returns 206 (Partial Content) when media is still processing
-            if flavour == .mastodon && httpResponse.statusCode == 206 {
-                return nil
-            }
-            // Re-throw for other status codes
-            throw TootSDKError.invalidStatusCode(data: data, response: httpResponse)
-        }
+        let response = try await getMediaRaw(id: id)
+        return response.data
     }
 
     /// Retrieve the details of a media attachment with HTTP response metadata
     ///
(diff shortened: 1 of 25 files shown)


Sources/TootSDK/TootClient/TootClient+Media.swift
```diff
@@ -42,6 +58,16 @@ extension TootClient {
     /// - Parameter id: The local ID of the attachment.
     /// - Returns: `Attachment` with a `url` to the media if available. `nil` otherwise.
     public func getMedia(id: String) async throws -> MediaAttachment? {
+        let response = try? await getMediaRaw(id: id)
```
Tunous, line 61: I don't think ignoring errors here is correct. A better approach would be to move the special mastodon case with status code 206 to this function. The raw function then could be simplified to use same logic like other api call without any special handling.

Tunous (commented):
👍 
I don't have a need for any headers so I won't comment on that part. Rest looks good for me.

davidgarywood (approved):
Looks good 🫡

kkostov:
Cool! Merging for now, if something comes up we will tackle it.

PR #361, opened 2025-08-02
Feature requirement can target fediverse server flavours by version

### LINKED ISSUE #359: Support API version requirement for feature flags
Currently, feature guards only check the flavour of the server e.g. mastodon, pleroma, akkoma etc. This doesn't account for the introduction of newer features as well as the deprecation of features in every flavour. 

In the current implementation, every feature declares a list of supported flavours: 

```swift
extension TootFeature {

    /// Ability to follow hashtags.
    ///
    public static let hashtagFollowing = TootFeature(supportedFlavours: [.mastodon, .friendica, .pixelfed])
}
```

This is [used by the](https://github.com/TootSDK/TootSDK/blob/bdcc62bb0d65b30f38acbc1c92d4d451de4d3175/Sources/TootSDK/TootClient/TootClient.swift#L263) `requireFeature` call to determine the availability of the feature: 

```swift
internal func requireFeature(_ feature: TootFeature) throws {
    try requireFlavour(feature.supportedFlavours)
}
```

Some background context: When TootSDK was first created, the Mastodon API had been mostly stable and other popular server flavours had similar of not identical implementation. As such, it was enough to only detect the server type.

Some examples of feature guards which require checking the specific flavour version: 

- Calling updateCredentials with the attributionDomains property is only supported if the server supports Mastodon API version 3 or higher. (#356)
- The endorse and unendorse endpoints are intended to replace the older pin and unpin which are now deprecated in Mastodon 4.4. (#356)
- Calling deleteMedia(id:) (#358)
- Using `deletePost` when calling deletePost(id:) (#358)

It seems we need a more comprehensive approach going forward.

## Solution

👉 We can extend `TootFeature` to accept (optional) from or range of versions for every flavour: 

```swift
public static let hashtagFollowing = TootFeature(supportedFlavours: [.mastodon(from: "2.0.0"), .friendica, .pixelfed])
```

or 

```swift
public static let hashtagFollowing = TootFeature(supportedFlavours: [.mastodon("1.2.3"..<"1.2.6"), .friendica, .pixelfed])
```

Here Version can be https://developer.apple.com/documentation/packagedescription/version (or similar) as long as it comes with comparison functionality and the range implementation is similar to the one we use to declare dependencies in [Swift Package Descriptions](https://developer.apple.com/documentation/packagedescription/package/dependency/package(id:_:)-27raa) 

👉 The `requireFeature` method can be extended to also check the version range if one is provided.

👉 Create a new struct e.g. `TootFlavourInfo` to hold the existing `TootSDKFlavour` enumeration as well as an optional associated value for the version number reported by the server.

```swift
struct TootFlavourInfo {
    let flavour: TootSDKFlavour
    let version: String?
}
```

In terms of usage, when a more complex check is required, we can implement some control flow:

```swift
public static let updateCredentialsAttributionDomains = TootFeature(supportedFlavours: [.mastodon(from: "4.4.0"), .friendica, .pixelfed])
```

```swift
func updateCredentials(params: UpdateCredentialsParams) async throws -> Account {
    // the flavour must support this feature, probably introduced in an older API
    try requireFeature(.updateCredentials)
    // Only Mastodon 4.4 and above support attribution domains:
    if let attributionDomains =  params.attributionDomains, !attributionDomains.isEmpty {
        try requireFeature(.updateCredentialsAttributionDomains)
    }
}
```

If the validation logic becomes too complex, we can also consider overloading requireFeature to accept the requirement + parameters so the specific toot method calls are more readable e.g.

```swift
func updateCredentials(params: UpdateCredentialsParams) async throws -> Account {
    try requireFeature(updateCredentials: params)

    // the rest of updateCredentials...
}

func requireFeature(updateCredentials: UpdateCredentialsParams) throws {
    // the flavour must support this feature, probably introduced in an older API
    try requireFeature(.updateCredentials)
    // Only Mastodon 4.4 and above support attribution domains:
    if let attributionDomains =  params.attributionDomains, !attributionDomains.isEmpty {
        try requireFeature(.updateCredentialsAttributionDomains)
    }
}

```

### Fallback behaviour

- If no server flavour version is detected, requireFeature calls without a version specification succeed.
- If no server flavour version is detected, requireFeature calls with a version specification fail.
- requireFeature calls without a version specification succeed as long as the server flavour is the expected one (the current behaviour today).
- requireFeature calls with version specification validate the required range against the detected server server version.

Feel free to share your thoughts on this.

kkostov:
Hi @daprice, this PR addresses the issue you previously raised regarding version checks (fixes #359). You are welcome to share any thoughts and comments!

The idea is that `TootFeature` is now responsible for parsing the flavour version and checking if it meets certain criteria. We now support several kinds of version checks depending on how the feature works (I updated `deleteMedia` as an example):

```swift
// Feature only for specific servers
let customFeature = TootFeature(supportedFlavours: [.mastodon, .pleroma])

// Feature with version requirements for specific servers
let versionedFeature = TootFeature(requirements: [
    .from(.mastodon, version: "4.0.0"),  // Mastodon 4.0+
    .any(.pleroma)                       // Any Pleroma version
])

// Feature supported by ALL servers, but with version requirements for some
let universalFeature = TootFeature(allExcept: [
    .from(.mastodon, version: "3.0.0"),  // Mastodon needs 3.0+
    .from(.pleroma, version: "2.0.0")    // Pleroma needs 2.0+
    // All other servers support any version
])

// Feature supported by specific servers, with version requirements for some
let selectiveFeature = TootFeature(
    anyExcept: [.friendica, .akkoma],    // These support any version
    versionRequirements: [
        .from(.mastodon, version: "3.5.0"),  // Mastodon needs 3.5+
        .from(.pixelfed, version: "2.0.0")   // Pixelfed needs 2.0+
    ]
)

// Check if current server supports it
if client.supportsFeature(customFeature) {
    // Use the feature
}
```

### DIFF
--- a/Sources/TootSDK/Models/TootFeature.swift
+++ b/Sources/TootSDK/Models/TootFeature.swift
@@ -1,8 +1,139 @@
 import Foundation
+import Version
 
 /// Represents a feature that is not supported by all flavours.
-public struct TootFeature {
+public struct TootFeature: Equatable {
 
-    /// Flavours that support this feature.
-    public let supportedFlavours: Set<TootSDKFlavour>
+    /// Represents a flavour requirement with optional version constraints
+    public struct FlavourRequirement: Equatable {
+        let flavour: TootSDKFlavour
+        let minVersion: Version?
+        let maxVersion: Version?
+
+        /// Create a requirement for any version of a flavour
+        public static func any(_ flavour: TootSDKFlavour) -> FlavourRequirement {
+            FlavourRequirement(flavour: flavour, minVersion: nil, maxVersion: nil)
+        }
+
+        /// Create a requirement for a minimum version
+        public static func from(_ flavour: TootSDKFlavour, version: String) -> FlavourRequirement {
+            FlavourRequirement(flavour: flavour, minVersion: Version(tolerant: version), maxVersion: nil)
+        }
+
+        /// Create a requirement for a version range
+        public static func range(_ flavour: TootSDKFlavour, from: String, to: String) -> FlavourRequirement {
+            FlavourRequirement(
+                flavour: flavour,
+                minVersion: Version(tolerant: from),
+                maxVersion: Version(tolerant: to)
+            )
+        }
+
+        /// Create a requirement for a maximum version
+        public static func until(_ flavour: TootSDKFlavour, version: String) -> FlavourRequirement {
+            FlavourRequirement(flavour: flavour, minVersion: nil, maxVersion: Version(tolerant: version))
+        }
+    }
+
+    /// The requirements for this feature
+    public let requirements: [FlavourRequirement]
+
+    /// Legacy initializer for backward compatibility
+    public init(supportedFlavours: Set<TootSDKFlavour>) {
+        self.requirements = supportedFlavours.map { FlavourRequirement.any($0) }
+    }
+
+    /// New version-aware initializer
+    public init(requirements: [FlavourRequirement]) {
+        self.requirements = requirements
+    }
+
+    /// Initialize with "any flavour except specific version requirements" pattern
+    /// - Parameters:
+    ///   - anyExcept: Flavours that should support any version
+    ///   - versionRequirements: Specific version requirements for certain flavours
+    public init(anyExcept: Set<TootSDKFlavour>, versionRequirements: [FlavourRequirement]) {
+        // Create requirements for "any version" flavours
+        let anyRequirements = anyExcept.map { FlavourRequirement.any($0) }
+        // Combine with specific version requirements
+        self.requirements = anyRequirements + versionRequirements
+    }
+
+    /// Initialize with default support for all flavours except specific version requirements
+    /// Useful for features that are generally available but have version constraints on certain servers
+    /// - Parameter versionRequirements: Specific version requirements for certain flavours
+    public init(allExcept versionRequirements: [FlavourRequirement]) {
+        // Get all flavours not mentioned in version requirements
+        let flavoursWithRequirements = Set(versionRequirements.map { $0.flavour })
+        let anyFlavours = Set(TootSDKFlavour.allCases).subtracting(flavoursWithRequirements)
+
+        // Create requirements for "any version" flavours
+        let anyRequirements = anyFlavours.map { FlavourRequirement.any($0) }
+        // Combine with specific version requirements
+        self.requirements = anyRequirements + versionRequirements
+    }
+
+    /// Check if feature is supported by an instance (for testing)
+    public func isSupported(by instance: any Instance) -> Bool {
+        return isSupported(flavour: instance.flavour, version: instance.version)
+    }
+
+    /// Check if feature is supported by a specific flavour and version string
+    public func isSupported(flavour: TootSDKFlavour, version: String?) -> Bool {
+        // Parse version with fallback to regex extraction
+        let versionObj = version.flatMap { Self.parseVersion(from: $0) }
+        return isSupported(flavour: flavour, version: versionObj)
+    }
+
+    /// Check if feature is supported by a specific flavour and parsed version
+    public func isSupported(flavour: TootSDKFlavour, version: Version?) -> Bool {
+        return requirements.contains { req in
+            guard req.flavour == flavour else { return false }
+
+            if let version = version {
+                if let minVersion = req.minVersion, version < minVersion {
+                    return false
+                }
+                if let maxVersion = req.maxVersion, version > maxVersion {
+                    return false
+                }
+            } else if req.minVersion != nil || req.maxVersion != nil {
+                // Version required but not available
+                return false
+            }
+
+            return true
+        }
+    }
+
+    /// Parse version from string with fallback to regex extraction
+    public static func parseVersion(from versionString: String) -> Version? {
+        // First try the Version library's tolerant parsing
+        if let version = Version(tolerant: versionString) {
+            return version
+        }
+
+        // If that fails, use regex to find the first version-like pattern
+        let pattern = #"\d+\.\d+(?:\.\d+)?"#
+        guard let regex = try? NSRegularExpression(pattern: pattern, options: []) else {
+            return nil
+        }
+
+        let range = NSRange(location: 0, length: versionString.utf16.count)
+        guard let match = regex.firstMatch(in: versionString, options: [], range: range) else {
+            return nil
+        }
+
+        guard let matchRange = Range(match.range, in: versionString) else {
+            return nil
+        }
+
+        let extractedVersion = String(versionString[matchRange])
+        return Version(tolerant: extractedVersion)
+    }
+
+    /// Legacy computed property for backward compatibility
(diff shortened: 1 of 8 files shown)


Sources/TootSDK/Models/TootFeature.swift
```diff
@@ -1,8 +1,139 @@
...
+    /// New version-aware initializer
+    public init(requirements: [FlavourRequirement]) {
+        self.requirements = requirements
+    }
+
+    /// Initialize with "any flavour except specific version requirements" pattern
+    /// - Parameters:
+    ///   - anyExcept: Flavours that should support any version
+    ///   - versionRequirements: Specific version requirements for certain flavours
+    public init(anyExcept: Set<TootSDKFlavour>, versionRequirements: [FlavourRequirement]) {
```
Tunous, line 55: This init is unclear to me. It reads as: support any flavor except these in `anyExcept`. I suggest `anyVersion` and `requirements` as parameter names.

Sources/TootSDK/TootClient/TootClient+Media.swift
daprice, line 1: Any flavor can declare support for Mastodon API v4 in their `instanceV2` response, not just Mastodon 4.4. 😀 I’d like to keep that distinction and base the TootFeature requirement on the API version, not the Mastodon software version. I wrote a reply in #361 with a longer version of my thoughts about this.

kkostov: Good point, I will adapt to look for InstanceV2 before applying the hard version requirement.

Sources/TootSDK/Models/TootFeature.swift
```diff
@@ -1,8 +1,303 @@
...
+        }
+
+        // MARK: - Combined Requirements
+
+        /// Create a requirement with both API version and display version fallback
+        /// - Parameters:
+        ///   - flavour: The server flavour
+        ///   - version: The minimum API version required
+        ///   - fallbackDisplayVersion: The minimum display version to use as fallback when API version is not available
+        public static func from(_ flavour: TootSDKFlavour, version: Int, fallbackDisplayVersion: String) -> FlavourRequirement {
```
Tunous, line 89: This one can be merged with one of previous by making fallback parameter optional. Less autocompletion suggestions/API to cover that way.

Sources/TootSDK/Models/TootFeature.swift
```diff
@@ -1,8 +1,303 @@
...
+                maxVersion: nil
+            )
+        }
+
+        // MARK: - Legacy Support (Deprecated)
+
+        /// Create a requirement for a version range (legacy method using display versions)
+        /// - Note: This method is deprecated. Use `from(_:displayVersion:to:)` for display versions or `from(_:version:to:)` for API versions
+        @available(*, deprecated, renamed: "from(_:displayVersion:to:)")
+        public static func range(_ flavour: TootSDKFlavour, from: String, to: String) -> FlavourRequirement {
```
Tunous, line 104: Is this needed? Version checks are a new feature. No one had a chance to use it.

Sources/TootSDK/Models/TootFeature.swift
```diff
@@ -1,8 +1,303 @@
...
+        // Combine with specific version requirements
+        self.requirements = anyRequirements + requirements
+    }
+
+    /// Deprecated initializer - use init(anyVersion:requirements:) instead
+    /// - Parameters:
+    ///   - anyExcept: Flavours that support any version (confusing parameter name)
+    ///   - versionRequirements: Specific version requirements for certain flavours
+    @available(*, deprecated, renamed: "init(anyVersion:requirements:)")
+    public init(anyExcept: Set<TootSDKFlavour>, versionRequirements: [FlavourRequirement]) {
```
Tunous, line 138: Also safe to remove instead of deprecating.

Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -458,11 +463,38 @@ extension TootClient {
 extension TootClient {
     /// Uses the currently available credentials to connect to an instance and detect the most compatible server flavour.
     public func connect() async throws {
-        if let flavour = await flavourFromNodeInfo() {
-            self.flavour = flavour
+        if let nodeInfo = await getNodeInfoIfAvailable() {
```
Tunous, line 466: This check looks problematic to me. The idea was that we prefer to use node info for flavour detection since it seemed to be the best place for that. Now that we also want to access apiVersions from instance v2 I think we should prefer that approach instead. Node info could be a fallback.

Sources/TootSDK/TootClient/TootClient.swift
```diff
@@ -458,11 +463,38 @@ extension TootClient {
...
+                }
+            }
+
+            if debugInstance {
+                print("🎨 Detected fediverse instance flavour: \(instance.flavour), version: \(instance.version)")
+            }
         }
+
+        // Set the encoder userInfo after flavour has been determined
+        encoder.userInfo[.tootSDKFlavour] = flavour
```
Tunous, line 493: I placed that in didSet of a flavour because it is public and can be modified by users of TootSDK. I do that myself when restoring from cache. Moving it here will break cases where someone overrides flavor manually. Overriding flavour/versions could be disallowed and users should be required to call connect or the old placement should be preserved.

I'm myself not yet sure how to update my caching approach to now work with version checks. I added it to avoid api calls since flavour never changed for a given instance.

Tunous (changes requested):

Sources/TootSDK/TootClient/TootClient.swift
kkostov: @Tunous That is a good point. I think the "server flavour" is now a combination of flavour and version info. I can still see cases where caching a TootClient may be useful, perhaps it's best to provide a dedicated initialiser for values that can be restored and make the actual properties get only.

Sources/TootSDK/Models/TootFeature.swift
```diff
@@ -1,8 +1,287 @@
...
+
+    /// Check if feature is supported considering API versions when available
+    /// - Parameters:
+    ///   - flavour: The server flavour
+    ///   - version: The parsed display version
+    ///   - apiVersions: The API versions from InstanceV2 (if available)
+    /// - Returns: true if the feature is supported
+    public func isSupported(flavour: TootSDKFlavour, version: Version?, apiVersions: InstanceV2.APIVersions?) -> Bool {
+        return requirements.contains { req in
+            guard req.flavour == flavour else { return false }
```
daprice, line 189: Thanks for taking my feedback into account, but I’m pretty sure this guard is still preventing the Mastodon API version from being checked for non-Mastodon server flavors.

kkostov: Ok, I just realised I didn't fully understand your previous comment and you're right for "cross-flavour" support checks, we need to also loo into apiVersions. Fix on the way! Thanks 🙏

Tunous (approved):

daprice (approved):
LGTM 👍🏻

PR #364, opened 2025-08-10
Include contentType in PostSource

Tunous:
In addition to the change reworked how test resources are located. The new solution avoids having to include each new file in the Package manifest.


Tunous:
@kkostov could this be merged and released before the other pr with version checks? I would like to release an app with this since it's small and leave version checks for next update.

kkostov:
@Tunous no problem, is there a reason you changed the way we load resource files for testing? I'd much rather keep them listed in Package.swift for now.

Tunous:
@kkostov for convenience. Having to remember to add them, make sure to not make typos and fight with Xcode not always picking up change was annoying. But I can revert that.

PR #368, opened 2025-08-22
Added an example of a TootSDK graphical client using SharingGRDB

thinkpractice:
This adds a basic example of a macOS gui app using `PointFree's SharingGRDB`. These are the changes:
- Removed all `SwiftData` related code
- Added [SharingGRDB](https://github.com/pointfreeco/sharing-grdb) framework to the package dependencies
- Reduced the macOS target to 15.0 (not on the latest version yet ;))
- Added `Models/Schema.swift` with the code to create the necessary tables & the database connection
- Adapted `Views/RootView.swift` to use `SharingGRDB` instead.
- Renamed some targets to reflect the new name.

All the code is in the `Examples/SharingGRDBExample` directory. Let me know if you have questions or if I need to change anything!

### DIFF
--- a/Examples/SharingGRDBExample/SharingGRDBExample.xcodeproj/project.pbxproj
+++ b/Examples/SharingGRDBExample/SharingGRDBExample.xcodeproj/project.pbxproj
@@ -1,0 +1,437 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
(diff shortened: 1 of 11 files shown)


kkostov (approved):
Thanks a lot for this example @thinkpractice! Really appreciated 🤩!

PR #376, opened 2025-09-05
Fix GoToSocial flavour detection

Tunous:
Fix for a regression introduced when moving to improved version detection. GoToSocial doesn't inform about its name in version string so the flavour detection based on instance didn't work and reported Mastodon.

Resolved by relying on information from node info. Only if it's not available information from instance will be used.


kkostov (approved):
Good catch!
I like the approach, making two requests here seems appropriate since node info lacks the `versions` object.

PR #384, opened 2025-10-03
add public init for PagedResult for better testability

omichde:
As I was adding pagination tests to my current project, I wanted to use `PagedResult` directly.
Sadly there was no public init for it, so swiftc was complaining.

- this adds the missing public init for `PagedResult` only
- no new test were added as this make the structure public only, no feature was changed
- existing tests all passed


kkostov (approved):
Awesome! Thank you!

PR #386, opened 2025-10-13
Fix unable to connect to services that do not support v2 instance info

Tunous:
The function to fetch instance info is called from connect initializer before flavour info is known. For that reason feature check for support of v2 instance succeeds because it thinks we are connecting to Mastodon.

To work around the issue also attempt to fetch v1 instance info if call to v2 endpoint fails with a http error.

___

This could be potentially further extended by setting flavour as soon as possible in the connect method but I decided to go for simplest fix for now.


kkostov (approved):

kkostov:
Thanks folks, that's a really good fix. ⛴️

PR #388, opened 2025-10-29
Support additional quote states from Mastodon 4.5

daprice:
See https://blog.joinmastodon.org/2025/10/mastodon-4-5-for-devs/

This PR doesn’t include quote authoring, only the additional states that quotes can be displayed as.

Thanks to PR #352 (OpenEnum), the new values coming from mastodon servers won’t break existing clients, so there’s no rush to update right away for this.


kkostov (approved):
It's awesome how things are coming together. Good thinking @daprice @Tunous 🤩 !

PR #400, opened 2026-04-22
Support block quotes in attributed string renderer

Tunous:
This pull request adds support for block quotes by rendering them with┃character prefix and in italics. Not 100% sure if this character is a good choice but I didn't find anything better. **I'm open for better suggestions.**

Additional small change is to prefix bullet points with a single space. It's a minor change but somehow without it the lists felt odd to me.

Example from my app how block quotes look like with this pull request
Simulator Screenshot - iPhone 17 - 2026-04-22 at 20 01 24

### DIFF
--- a/Examples/TootSDKExample/TootSDKExample.xcodeproj/project.pbxproj
+++ b/Examples/TootSDKExample/TootSDKExample.xcodeproj/project.pbxproj
@@ -7,7 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		B5362EB92E54C1790075BBD2 /* TootSDK in Frameworks */ = {isa = PBXBuildFile; productRef = B5362EB82E54C1790075BBD2 /* TootSDK */; };
+		B51DCC1C2F9A761D00AFB55B /* TootSDK in Frameworks */ = {isa = PBXBuildFile; productRef = B51DCC1B2F9A761D00AFB55B /* TootSDK */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -69,7 +69,7 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				B5362EB92E54C1790075BBD2 /* TootSDK in Frameworks */,
+				B51DCC1C2F9A761D00AFB55B /* TootSDK in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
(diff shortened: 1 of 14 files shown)


kkostov:
Support for block quotes is an excellent idea, thank you for putting together this PR @Tunous!

Renderers are something we assume folks would just fork and create their own as they probably want things a bit differently in their apps.

With that said, maybe we can implement a locale-aware solution for the built-in renderer.

As it happens, I've been reading a lot about typography recently 😇 and especially around how quotation marks and glyphs render in text and how they change depending on locale or context (there are soo many quirks in different languages!)


For example, a "plain text" citation could look like:

```
// en: “That's one way to do it”
// fr: « That's one way to do it »
// de: „That's one way to do it"
// ja: 「That's one way to do it」
```


To get to that, we could take advantage of [Locale.quotationBeginDelimiter](https://developer.apple.com/documentation/foundation/locale/quotationbegindelimiter) and [Locale.alternateQuotationBeginDelimiter](https://developer.apple.com/documentation/foundation/locale/alternatequotationbegindelimiter) for first and next level quotes as they're configured by the system. In case of French or Italian, we can add the expected spaces too (\u{00A0}).

Users of the library could pass some styling configuration to the renderer (notice how we can style the glyph and the content of the quotes separately): 

```swift
let blockQuoteStyle: BlockQuoteStyle = {
        var marks = AttributeContainer()
        marks.swiftUI.font = .system(size: 32, weight: .bold)
        marks.swiftUI.foregroundColor = .green

        var body = AttributeContainer()
        body.swiftUI.font = .body

        return BlockQuoteStyle(
            locale: Locale(identifier: "fr_FR"),
            markAttributes: marks,
            contentAttributes: body
        )
    }()
```

```
public func render(html: String, options: Options = [], blockQuoteStyle: BlockQuoteStyle = .default) -> ParsedContent
```

Then we get something like this: 

Screenshot 2026-04-23 at 18 52 43

Screenshot 2026-04-23 at 18 56 00

(for English and French respectively based on the locale value)

I tried to play with it some [on a separate branch](https://github.com/TootSDK/TootSDK/blob/feat/quote-experiments/Examples/TootSDKExample/TootSDKExample/Views/MentionsTimelineView.swift#L24) - feel free to check it out and take what you like into this PR! And of course if you have other ideas, feel free to share!

Tunous:
Great solution! That's much better than what I had. I went ahead and merged your changes to this branch. They work for me and I don't see any issues. I managed to come up with a simple style that fits my app.

Simulator Screenshot - iPhone 17 - 2026-04-23 at 19 48 48

kkostov (approved):
Awesome!

PR #402, opened 2026-05-30
Account avatar description property

Tunous:
https://docs.joinmastodon.org/entities/Account/#avatar_description


kkostov (approved):
Awesome! Thank you @Tunous

PR #407, opened 2026-07-22
Add dedicated parameters for rescheduling posts

### LINKED ISSUE #267: excess params for updateScheduledPostDate
`updateScheduledPostDate` takes `ScheduledPostParams` which has all the info for creating a scheduled post, but the update only uses the schedule_at date

https://docs.joinmastodon.org/methods/scheduled_statuses/#update

currently all the other param fields can be omitted in a call except the post visibility since it's not optional, so I call it like this

```swift
 try await updateScheduledPostDate(
      id: post.id,
      ScheduledPostParams(
        visibility: .direct,  // kludge, not used
        scheduledAt: date
      ))
```

easy workaround but it probably should be revised with a new param struct like ReschedulePostParams and the doc clarified (and code could be simplified if you don't need to do that multipart thing)

almalehb:
This closes #267

`updateScheduledPostDate` currently accepts `ScheduledPostParams` which includes unrelated parameters like visibility even though the endpoint only needs `scheduled_at`

This PR adds a smaller `ReschedulePostParams` type containing only `scheduledAt` and updates the method to use URL-form encoding

- preserved existing five-minute date validation
- adds `updateScheduledPostDateRaw`
- deprecates the existing overload without breaking current callers
- adds tests for form encoding and date validation

### DIFF
--- a/Sources/TootSDK/Models/ReschedulePostParams.swift
+++ b/Sources/TootSDK/Models/ReschedulePostParams.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+/// Parameters to update a scheduled post's publication date.
+public struct ReschedulePostParams: Sendable {
+
+    /// Date and time at which the post will be published.
+    public let scheduledAt: Date
+
+    /// Creates parameters to update a scheduled post's publication date.
+    ///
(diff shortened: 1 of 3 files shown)


kkostov (approved):

kkostov:
Awesome, thanks a lot @almalehb 🎊 !

